PR #8, opened 2018-09-10
Only emit node:ready when p2p is for sure ready

harbu:
Moved libp2p ready check to earlier phase so event `node:ready` is only emitted when libp2p is really ready.

### DIFF
--- a/src/connection/Connection.js
+++ b/src/connection/Connection.js
@@ -51,7 +51,11 @@ module.exports = class Connection extends EventEmitter {
             debug('node has started: %s', this.node.isStarted())
             this.isStarted = this.node.isStarted()
 
-            this.emit('node:ready')
+            if (!this.isStarted) {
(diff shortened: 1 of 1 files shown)


src/connection/Connection.js
```diff
@@ -51,7 +51,11 @@ module.exports = class Connection extends EventEmitter {
             debug('node has started: %s', this.node.isStarted())
             this.isStarted = this.node.isStarted()
 
-            this.emit('node:ready')
+            if (!this.isStarted) {
```
mirpo: we don't need this check, check this function https://github.com/libp2p/js-libp2p/blob/master/src/index.js#L234
and actually can use original _isStarted from libp2p https://github.com/libp2p/js-libp2p/blob/master/src/index.js#L39

harbu:
Removed the check, better @mirpo ?

mirpo (approved):

PR #28, opened 2018-09-11
Eslint + fixes to style

harbu:
Took Eslint config from https://github.com/streamr-dev/eslint-config-streamr but removed JSX and Babel stuff. Then took some options from https://github.com/streamr-dev/data-api/blob/master/.eslintrc.js.

Also did some additional tuning:
- `'arrow-body-style': DISABLED` (wanted to force me to make utils.js#callbackToPromise into a very long one-liner which I found less readable.)
- ` 'comma-dangle': DISABLED` (meh, don't think this is so important)
- `max-len` from `ERROR` to `WARN` (we have the very long private key as constant var in tracker.js)
- `'no-unused-vars': WARN`
- `'prefer-template': DISABLED` (Otherwise forces to use string templates. I think sometimes concatenation actually looks nicer, depends on situation.)

I didn't touch tests yet still @mirpo working on those atm, and trying to avoid silly git conflicts.


node.js
```diff
@@ -1,12 +1,12 @@
 const Connection = require('./src/connection/Connection')
 const Node = require('./src/logic/Node')
+
```
harbu, line 3: Wants line break after requires are introduced.

package.json
```diff
@@ -40,14 +40,16 @@
     "libp2p-tcp": "^0.12.1",
     "libp2p-websockets": "^0.12.0",
     "ms": "^2.1.1",
+    "peer-id": "^0.11.0",
```
harbu, line 43: ESLint doesn't like if you require a library that isn't mentioned here.

src/connection/Connection.js
```diff
@@ -20,6 +18,10 @@ const events = Object.freeze({
     MESSAGE_RECEIVED: 'streamr:message-received'
 })
 
+async function getPeerInfo(conn) {
```
harbu, line 21: ESLint wanted function to be moved here from inside the class because `this` is not referenced.

src/connection/Connection.js
```diff
@@ -86,10 +84,8 @@ module.exports = class Connection extends EventEmitter {
         })
         this.node.on('peer:disconnect', (peer) => this.emit(events.PEER_DISCONNECTED, peer))
 
-        this.on('streamr:send-message', ({
-            recipient,
-            message
-        }) => this._onSend(recipient, message))
+        this.on('streamr:send-message', ({ recipient,
```
harbu, line 87: Current ESLint config doesn't allow spread operator opening bracket to be followed by a newline.

src/connection/Connection.js
```diff
@@ -116,36 +115,29 @@ module.exports = class Connection extends EventEmitter {
...
         try {
-            const sender = await this._getPeerInfo(conn)
+            const sender = await getPeerInfo(conn)
 
             pull(
                 conn,
-                pull.map(message => message.toString('utf8')),
-                pull.drain(message => {
-
+                pull.map((message) => message.toString('utf8')),
```
harbu, line 122: ESLint wants "lambda" function parameter list to always be surrounded by parentheses

src/connection/Connection.js
mirpo, line 21: why it can't be a class method?

src/helpers/TopologyStrategy.js
```diff
@@ -1,8 +1,8 @@
-const getPeersTopology = (peers, currentPeer) => _getRandomPeers(_filterOutPeer(peers, currentPeer))
+const getRandomPeers = (peers, max = 3) => peers
```
harbu, line 1: ESLint requires function to be declared before it is used in another function

src/protocol/TrackerNode.js
```diff
@@ -36,27 +30,21 @@ module.exports = class TrackerNode extends EventEmitter {
...
     onReceive(sender, message) {
-        const {
-            code,
-            data
-        } = encoder.decode(message)
+        const { code, data } = encoder.decode(message)
 
         switch (code) {
             case encoder.PEERS:
-                const peers = data
```
harbu, line 46: ESLint doesn't like variables to be declared inside switch-case

src/logic/Node.js
```diff
@@ -28,16 +24,9 @@ module.exports = class Node extends EventEmitter {
 
         this.connection.once('node:ready', () => this.onNodeReady())
         this.listners.trackerNodeListner.on('streamr:peer:send-status', (tracker) => this.onSendStatusToTracker(tracker))
-        this.listners.trackerNodeListner.on('streamr:node-node:stream-data', ({
-            streamId,
-            data
-        }) => this.onSendData(streamId, data))
+        this.listners.trackerNodeListner.on('streamr:node-node:stream-data', ({ streamId, data }) => this.onSendData(streamId, data))
```
mirpo, line 27: you ran style update not on the latest version with event names?

src/connection/Connection.js
harbu: I think because it doesn't use `this`. Current ESLint rules doesn't allow that.

mirpo (approved):
I don't agree with moving getPeerInfo outside of the class, but everything else is ok

harbu:
I don't really have strong feelings either way. I sort of see the reasoning ESLint+Airbnb has: "only methods that access `this` should be methods" but then again it kind of results in weird code where you need to have functions outside the class. Dunno. 

We can change the eslint rules if you want.

src/connection/Connection.js
mondoreale: Original `_getPeerInfo` was an instance method, not a class method. Eslint says if it's an instance method, make it relative to the instance (use `this` in it at least once). Otherwise it does not need to be an instance method but a helper/utility. You can still make it a class method tho:
```jsx
class Connection extends Whatever {
  static async getPeerInfo = (conn) => {
    // …
  }
}
```
Makes sense if you want to use it outside of this file, e.g. 
```jsx
import Connection from './connection'

const info = Connection.getPeerInfo(lovelyConnection)
```
Cheers!

harbu: Thanks we will try it out!

Issue #30, opened 2018-09-11
Send messages from Logic layer via Protocol layer


harbu:
I guess the idea is that logic layer wouldn't send messages directly via `connection` object, but instead use an interface provided by protocol layer to do this.

ptesavol:
Exactly like this. Actually the logic layer should not even know about the existence of the connection layer. The composition of the layers can be done at the "executable layer" (in /tracker.js and similar files)

PR #32, opened 2018-09-11
Communicate from Logic Layer via Protocol Layer

harbu:
Changed code so that sending messages from Logic layer happens via interfaces provided by Protocol layer instead of Connection layer directly.

#30


ptesavol (approved):
Work is still needed to remove all unnecessary coupling between the layers, but this is a good step to that direction. Next thing to do will be moving the layer composition away from the logic layer into the executable layer.

mirpo (approved):

PR #38, opened 2018-09-12
Add draft unit and integration tests, code coverage is around 75%

mirpo:
fixes #14

PR #46, opened 2018-09-12
Export static events

### LINKED ISSUE #36: Export events from connection and protocols
to use them across the project

harbu:
fixes #36 

- Exported event constants from protocol layer and connection layer so that higher levels can use them. 
- Refer to events in higher levels by their constant instead of string name.

## TODO
- Integrate with #38


mirpo (approved):
looks good, I'll update tests to use those events too

PR #54, opened 2018-09-14
Stream responsibility

### LINKED ISSUE #51: Assigning stream responsibility
- First node asks tracker who is responsible for stream.
- If no one is responsible, then tracker assigns responsibility to this node.

mirpo:
fixes #51


src/logic/Node.js
```diff
@@ -52,32 +46,44 @@ class Node extends EventEmitter {
...
+    isOwnStream(streamId) {
+        return this.ownStreams.has(streamId)
     }
 
     _isKnownStream(streamId) {
         return this.knownStreams.get(streamId) !== undefined
     }
+
+    stop(cb) {
+        this.protocols.nodeToNode.stop(cb)
```
harbu, line 85: Should we stop `trackerNode` also?

src/protocol/TrackerNode.js
```diff
@@ -8,7 +8,8 @@ const events = Object.freeze({
     CONNECTED_TO_TRACKER: 'streamr:peer:send-status',
     NODE_LIST_RECEIVED: 'streamr:node-node:connect',
     DATA_RECEIVED: 'streamr:node-node:stream-data',
-    STREAM_INFO_RECEIVED: 'streamr:node:found-stream'
+    STREAM_INFO_RECEIVED: 'streamr:node:found-stream',
+    STREAM_INFO_OWN_RECEIVED: 'streamr:node:found-own-stream',
```
harbu: Hmmm.. what about name `STREAM_ASSIGNED`?

harbu (commented):
Looks good to me. Few comments; but nothing critical. Feel free to merge if you want.

src/logic/Node.js
mirpo: hmmm, we are using one connection, so we don't need to stop all protocols, we only need to stop libp2p connection

src/protocol/TrackerNode.js
mirpo: good point

Issue #34, opened 2018-09-11
Move composition of the layer stack away from the logic layer


ptesavol:
The composition of the layer stack needs to be moved away from the logic layer into the executable layer (into /tracker.js etc.). Logic layer should not need to know about the existence of the connection layer.

PR #57, opened 2018-09-15
Move composition of the layer stack away from the logic layer (#34)

ptesavol:
This is only the first step. Now composition of the layers is done in many places: in the executable and the tests. What do you think about encapsulating the layer composition info factories such as TrackerFactory and NodeFactory that would be the only places where the composition is done?


mirpo:
please update code, so it'll pass Travis -> eslint + existing tests

harbu (changes requested):
LGTM; but yeah travis should pass before merge

publisher.js
```diff
@@ -8,7 +9,7 @@ const streamId = process.argv[4] || ''
 createConnection('127.0.0.1', port, '', true).then((connection) => {
     connection.connect(nodeAddress)
 
-    const publisher = new Publisher(connection, nodeAddress)
+    const publisher = new Publisher(new NodeToNode(connection), nodeAddress)
```
mirpo, line 12: Maybe it's  better to something like `async createConnection`? https://github.com/streamr-dev/network/blob/master/src/connection/Connection.js#L143
Or move it to some helper file, so we can reuse it in code, tests and etc.

src/NetworkNode.js
```diff
@@ -41,6 +43,6 @@ class NetworkNode extends EventEmitter {
 
 module.exports = async (host, port, key = '') => {
     const connection = await createConnection(host, port, key, true)
-    const node = new Node(connection)
+    const node = new Node(new TrackerNode(connection), new NodeToNode(connection))
```
mirpo, line 46: or something like Node({trackerNode: new TrackerNode(connection), nodeToNode: new NodeToNode(connection)})
so in constructor only this.protocols =

src/logic/Node.js
```diff
@@ -22,11 +22,10 @@ class Node extends EventEmitter {
...
 
         this.protocols = {
-            trackerNode: new TrackerNode(connection),
-            nodeToNode: new NodeToNode(connection)
+            trackerNode: trackerNode,
+            nodeToNode: nodeToNode
         }
 
-        connection.once('node:ready', () => this.onNodeReady())
         this.protocols.trackerNode.on(TrackerNode.events.CONNECTED_TO_TRACKER, (tracker) => this.onConnectedToTracker(tracker))
```
mirpo: we must somehow validate that we have correct protocols, which will catch events.

mirpo (changes requested):

publisher.js
ptesavol: How about adding folder called "composition" and putting there classes TrackerFactory and NodeFactory for creating the layer composition of each. It would then be easy to change the composition to something else, for example replace the libp2p connection layer with plain websockets connection layer.

src/logic/Node.js
ptesavol: I searched the file and there was no such method as onNodeReady(), that is why I removed the code that calls that function.

publisher.js
ptesavol: I'll make a separate issue about this and do this next

harbu (approved):

src/logic/Node.js
harbu: Yeah I got rid of the event in a previous commit.

mirpo (approved):

PR #63, opened 2018-09-17
update terminology in README

### LINKED ISSUE #44: Domain language: Pub/Sub vs Producer/Consumer
We should lock down terminology for consistency of terms in codebase.

Do we use:
Publisher / Subscriber

or

Producer / Consumer

or

something else?

harbu:
Define terms client, publisher, subscriber, and also broker and stream.

fixes #44, #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ layer and pushes new data to the Streamr network via the same layer.
 
 # Glossary
 - A _peer_ is any participant in the peer-to-peer network.
-- A _tracker_ is a peer that assists nodes to discover other nodes.
+- A _tracker_ assists nodes to discover other nodes. Peers can act as trackers, but a tracker might also be a centralised server in some configurations
 - A _node_ is a peer that forwards data in the Streamr network pub-sub.
 - A _broker_ is a node that also includes client-facing functionality and interfaces.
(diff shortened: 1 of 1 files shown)


ptesavol (changes requested):
I would say "A _tracker_ assists nodes to discover other nodes. Peers can act as trackers but a tracker might also be a centralised server in some configurations".

In the two last lines I would replace word "node" with "broker".

How about also adding definition of _leader_?

harbu:
I replaced word _node_ with _broker_ in the definitions of _publisher_ and _subscriber_. With regards to the wording of _tracker_, does the statement "Peers can act as trackers but a tracker might also be a centralised server in some configurations" apply in our case?

ptesavol:
I think in some private, commercial setups (such as the smart city case) a centralised tracker may be relevant.

harbu:
Fair enough, updated definition.

mirpo (approved):
what about code? should we rename all classes after that?

harbu:
If time allows, I think we should make a pass through the code to make sure stuff is named appropriately. But most of all, we should make sure to use follow these terms in future code we write.

harbu:
Whoops accidentally closed, re-opened...

ptesavol (approved):

PR #65, opened 2018-09-18
Composition layer and parameterized bootstrap nodes 

### LINKED ISSUE #58: Move composition of layer stack to factory classes
The composition of the layer stack should be done in one place so it is easy to make changes to the composition.

harbu:
fixes #58 

I isolated these changes from the upcoming messaging test I'm building because it still requires some work.
 
- Composition layer: `composition.js` contains functions for starting nodes, trackers, and network nodes.
- Allow bootstrap nodes to be given via argument list.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,0 @@
-const startNetworkNode = require('./src/NetworkNode')
-
-module.exports = {
-    startNetworkNode
-}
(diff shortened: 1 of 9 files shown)


ptesavol (approved):
Looks good. How about using composition.js also in the tests?

mirpo (approved):
need to update tests

PR #68, opened 2018-09-18
New status

### LINKED ISSUE #55: Inform tracker about new status
after #51

### LINKED ISSUE #61: Remove all text events from tests

mirpo:
- fixes #61
- fixed #55 
- update logic of requesting new peers from tracker

### DIFF
--- a/src/logic/Tracker.js
+++ b/src/logic/Tracker.js
@@ -32,6 +32,9 @@ module.exports = class Tracker extends EventEmitter {
             debug('sending list of nodes')
             this.protocols.trackerServer.sendNodeList(node, listOfNodes)
         }
+        else {
+            debug('no available nodes to send')
+        }
     }
 
     processNodeStatus(node, status) {
--- a/src/logic/Tracker.js
+++ b/src/logic/Tracker.js
@@ -26,10 +26,12 @@ module.exports = class Tracker extends EventEmitter {
     }
 
     sendListOfNodes(node) {
-        debug('sending list of nodes')
-
         const listOfNodes = getPeersTopology([...this.nodes.keys()], getAddress(node))
-        this.protocols.trackerServer.sendNodeList(node, listOfNodes)
+
+        if (listOfNodes.length) {
+            debug('sending list of nodes')
+            this.protocols.trackerServer.sendNodeList(node, listOfNodes)
+        }
     }
 
     processNodeStatus(node, status) {
--- a/src/logic/Tracker.js
+++ b/src/logic/Tracker.js
@@ -31,8 +31,7 @@ module.exports = class Tracker extends EventEmitter {
         if (listOfNodes.length) {
             debug('sending list of nodes')
             this.protocols.trackerServer.sendNodeList(node, listOfNodes)
-        }
-        else {
+        } else {
             debug('no available nodes to send')
         }
     }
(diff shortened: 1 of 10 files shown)


src/logic/Tracker.js
```diff
@@ -26,10 +26,12 @@ module.exports = class Tracker extends EventEmitter {
...
     sendListOfNodes(node) {
-        debug('sending list of nodes')
-
         const listOfNodes = getPeersTopology([...this.nodes.keys()], getAddress(node))
-        this.protocols.trackerServer.sendNodeList(node, listOfNodes)
+
+        if (listOfNodes.length) {
+            debug('sending list of nodes')
+            this.protocols.trackerServer.sendNodeList(node, listOfNodes)
+        }
```
harbu: Should we do a `debug` message if there are no nodes to send? To make sure that Tracker is operating correctly.

src/protocol/TrackerNode.js
```diff
@@ -43,6 +45,14 @@ class TrackerNode extends EventEmitter {
         }
     }
 
+    requestMorePeers() {
+        if (this.peersInterval === null) {
+            this.peersInterval = setInterval(() => {
+                this.connection.send(this.tracker, encoder.peersMessage([]))
+            }, 10000)
```
harbu: On 1st call it will wait 10 seconds to fetch peers? Kind of a long wait? Why can't we immediately request more peers?

test/integration/connection.test.js
```diff
@@ -1,7 +1,7 @@
 const assert = require('assert')
-const { getTestConnections } = require('../util')
+const { getTestConnections, DEFAULT_TIMEOUT } = require('../util')
 
-jest.setTimeout(60000)
+jest.setTimeout(DEFAULT_TIMEOUT)
```
harbu, line 4: I think it is actually good idea to have timeout defined per test. Because some tests we expect to be slow, and others we expect to be fast. If we know that a test should be fast, and it isn't, it is probably a failure and we want it to fail fast instead of having to wait.

test/util.js
```diff
@@ -1,12 +1,14 @@
-const localhost = '127.0.0.1'
+const LOCALHOST = '127.0.0.1'
```
harbu: I think `require`(s) should be first in file, followed by constants.

harbu (changes requested):
Looks good overall. Few points, especially the one related to `setInterval`

src/protocol/TrackerNode.js
mirpo: this situations happens only when we have only one tracker and one node, so I'll update it for 1sec

src/logic/Tracker.js
mirpo: agree

test/integration/connection.test.js
mirpo: I added timeouts only to slow tests, or you mean to have different timeouts for different tests?

test/util.js
mirpo: ok

test/integration/connection.test.js
harbu: I think its okay then if its only slow tests! 👍

harbu (approved):

PR #73, opened 2018-09-18
speed up travis

harbu:
Make Travis CI faster by caching `node_modules`. This avoids having to `npm install` from scratch on every run.


harbu:
For some reason, makes things slower... Forget about this now.

Issue #67, opened 2018-09-18
Streamline protocol layer classes


ptesavol:
The three protocol layer classes look quite different from each other even though they should all be doing the same thing in the southbound side: listening to the same events from the layer below, parsing them and generating events to the upper layer. 

- All events generated by the connection layer should be listened to in all 3 classes
- Southbound event listener methods should be named the same in all 3 classes  

To implement this, I propose implementing the UML diagram's EndpointListener interface as a real class that would do/enforce this all in one place.

PR #79, opened 2018-09-20
Streamline protocol layer classes, issue #67

ptesavol:
Implemented EndpointListener interface as a Javascript class, and made all protocol classes to implement this interface. This removes duplicated code (importing and binding Endpoint events) from the protocol classes, and makes sure that the protocol classes listen to all the events coming from the lower layer.

The "implement(implementor, endpoint)" method of EndpointListener class checks whether the implementor has actually implemented the interface's methods and throws an exception otherwise. If the methods exist as claimed, it binds the corresponding endpoint events to the methods.

Jslint config was slightly modified to only give a warning when empty placeholder methods are encountered. I also removed the warning about exceeding the line length, which happens in almost all the files in the project.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -56,7 +56,7 @@ module.exports = exports = {
             numbers: true
         }],
         'semi': [ERROR, 'never'],
-        'max-len': DISABLED,
+        'max-len': WARN,
         'class-methods-use-this': DISABLED,
         'no-empty-function': WARN,
     }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -56,5 +56,8 @@ module.exports = exports = {
             numbers: true
         }],
         'semi': [ERROR, 'never'],
+        'max-len': DISABLED,
+        'class-methods-use-this': DISABLED,
+        'no-empty-function': WARN,
     }
 }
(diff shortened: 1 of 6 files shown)


mirpo:
what do you think about https://www.npmjs.com/package/implement-js ?

mirpo (approved):

ptesavol:
Implement-js looks promising in the sense that it is run-time (as oppesed to flowtype). However there seems to be an open issue concerning es6 classes: https://github.com/Jahans3/Implement.js/issues/6

.eslintrc.js
```diff
@@ -56,5 +56,8 @@ module.exports = exports = {
             numbers: true
         }],
         'semi': [ERROR, 'never'],
+        'max-len': DISABLED,
```
harbu: Can we still leave this at `WARN`? We have it currently set up at 150 char limit, and that is already quite a long line.

harbu (changes requested):
Looks good to me, but I think we should leave the max-len to WARN.

.eslintrc.js
ptesavol: yep, I'll make the change!

harbu (approved):

PR #83, opened 2018-09-21
Speedup Travis build

kare:


kare:
Container is the default after all

Issue #85, opened 2018-09-21
Buffering of messges


harbu:
First message send to node fails if node doesn't know who is leader (responsible) for that stream. Node should buffer message until this is discovered. Maybe implement timeout, or leave for later date.

PR #87, opened 2018-09-21
logic/Node: buffer messages with unknown streamId

harbu:
When a Node receives a message with a stream it hasn't seen before, it will buffer that message for later sending. Two things can occur hence after.  

1. Node receives information of who is responsible for the stream (or it is assigned that responsibility). It will then proceed to send all messages of that stream from the buffer (or in the case where it itself is responsible, it will emit events to listeners.)
2. Node fails to deliver messages in given timeframe (60 seconds currently), after which it emits a `MESSAGE_DELIVERY_FAILED` event that users can react to.

Also included is an integration test verifying that initial messages are delivered.

(Fixes #85)


mirpo (approved):
looks really good

PR #92, opened 2018-09-24
Message numbering

harbu:
## Changes
- Messages now have `number` and `previousNumber`. Later refers to number of previous message, or null if is 1st message in stream.
- Leader of stream numbers messages
- Message numbers are managed by `StreamManager`
- When receiving message for other leader's stream
   - if it has number, pass to subscribers
   - if not, pass to its leader for numbering 
- `Node#_sendToSubscribers`: only emit `MESSAGE_RECEIVED` event if we have subscription to stream.
- Catch and log invalid message numberings (e.g. `previousNumber != null && number == null` or leader receiving already numbered message)
- Cleaned up some debugging messages

## Known limitations
- No ability to store message numbers: if leader drops and new one joins, message numbering starts from `1` again.
- No duplicate detection or gap detection currently for Nodes
- Clients have no ability to request missing messages

## Motivation
Leader nodes now assign messages (of streams that they are leaders of) with a `number` from a strictly increasing integer sequence (per stream). Numbering provide _identity_ to messages within a stream, allowing nodes and clients to detect whether a message has been seen before or not, and thus whether to skip it. Numbering also provides _order_ within a stream of messages. This allows clients and nodes to detect gaps (missing messages in-between that have not been received) and process messages in the intended order. It also lessens space requirements for duplicate detection (don't have to keep track of every single number in memory, but can instead rely on a lower bound.)


mirpo (changes requested):
first merge master, reviewing code with previous version

mirpo (approved):

PR #93, opened 2018-09-25
test: faster travis

harbu:


harbu:
Not good because cannot differentiate between test types.

Issue #82, opened 2018-09-21
Message forwarding


harbu:
Forward messages to also other nodes that have subscribed to a stream.

PR #94, opened 2018-09-25
Subscriber implementation

### LINKED ISSUE #88: encoder.DATA should be in NodeToNode
This https://github.com/streamr-dev/network/blob/master/src/protocol/TrackerNode.js#L71 should probably be in `NodeToNode` instead of `TrackerNode`.

mirpo:
@harbu I only commented nodeMessageBuffering.test.js, need your help to fix it
fixes #88, #94, #82


src/logic/Client.js
```diff
@@ -12,19 +12,28 @@ module.exports = class Publisher extends EventEmitter {
...
             throw new Error('Failed to publish because this.nodeAddress not defined.')
         }
     }
 
+    subscribe(streamId) {
+        if (this.nodeAddress) {
+            this.debug('subscribing to stream %s', streamId)
+            this.protocols.nodeToNode.sendSubscribe(this.nodeAddress, streamId)
+        } else {
+            throw new Error('Failed to subscribe because this.nodeAddress not defined.')
```
harbu, line 33: Would it make sense to already throw in constructor to fail early?

src/logic/Node.js
```diff
@@ -17,6 +17,7 @@ class Node extends EventEmitter {
 
         this.knownStreams = new Map()
         this.ownStreams = new Set()
+        this.subsribers = new Map()
```
harbu: rename subsribers -> subscribers

src/logic/Node.js
```diff
@@ -84,6 +93,49 @@ class Node extends EventEmitter {
...
+        this.debug('node %s unsubscribed from the stream %s', nodeAddress, streamId)
+        this._unsubscribeNode(streamId, nodeAddress)
+    }
+
+    _unsubscribeNode(streamId, nodeAddress) {
+        if (this.subsribers.has(streamId)) {
+            this.subsribers.set(streamId, [...this.subsribers.get(streamId)].filter((node) => node !== nodeAddress))
+        }
+
+        // check and unsubscribe from that stream
```
harbu: What about this? Separate PR?

src/logic/Node.js
```diff
@@ -99,6 +151,26 @@ class Node extends EventEmitter {
         this.protocols.nodeToNode.stop(cb)
     }
 
+    _addToSubscribers(streamId, nodeAddress) {
+        if (this._checkPermissions(streamId, nodeAddress)) {
```
harbu, line 155: Hmm.. I'm still not convinced we need permission checking at the network level. But let us discuss this with Henri and Petri in next meeting.

src/protocol/NodeToNode.js
```diff
@@ -38,8 +52,42 @@ module.exports = class NodeToNode extends EventEmitter {
...
+                break
+
+            case encoder.UNSUBSCRIBE:
+                this.emit(events.UNSUBSCRIBE_REQUEST, {
+                    streamId: data,
+                    sender
+                })
+                break
+
+            case encoder.DATA:
```
harbu, line 72: 👍

src/protocol/TrackerNode.js
```diff
@@ -87,7 +80,7 @@ class TrackerNode extends EventEmitter {
                 break
 
             default:
-                throw new Error('Unhandled message type')
+                break
```
harbu, line 83: Hmm.. can we at least debug here? If there are unexpected messages that is a sign of a bug and would be good that we know about it.

src/protocol/TrackerServer.js
```diff
@@ -66,12 +70,16 @@ class TrackerServer extends EventEmitter {
                 break
 
             default:
-                throw new Error('Unhandled message type')
+                break
```
harbu, line 73: Same comment as above

src/protocol/NodeToNode.js
```diff
@@ -38,8 +52,42 @@ module.exports = class NodeToNode extends EventEmitter {
...
+                this.emit(events.DATA_RECEIVED, {
+                    streamId: data[0],
+                    data: data[1]
+                })
+                break
+
+            default:
+                break
```
harbu, line 80: Same comment as below.

subscriber.js
```diff
@@ -0,0 +1,27 @@
+const { createEndpoint } = require('./src/connection/Libp2pEndpoint')
+const Client = require('./src/logic/Client')
+const NodeToNode = require('./src/protocol/NodeToNode')
+
+const port = process.argv[2] || 30301
+const nodeAddress = process.argv[3] || ''
+const streamIdParam = process.argv[4] || ''
+
+createEndpoint('127.0.0.1', port, '', true).then((endpoint) => {
```
harbu, line 9: I think you could use `startClient` here?

test/integration/nodeMessageBuffering.test.js
```diff
@@ -1,9 +1,11 @@
 const { startNode, startTracker } = require('../../src/composition')
 const Node = require('../../src/logic/Node')
 const { callbackToPromise } = require('../../src/util')
-const { PRIVATE_KEY } = require('../util')
+const { waitForEvent, LOCALHOST, DEFAULT_TIMEOUT, PRIVATE_KEY } = require('../util')
+const TrackerNode = require('../../src/protocol/TrackerNode')
+const TrackerServer = require('../../src/protocol/TrackerServer')
 
-jest.setTimeout(30 * 1000)
+jest.setTimeout(DEFAULT_TIMEOUT)
```
harbu, line 8: Let us have a look at this.

test/integration/subscriber.test.js
```diff
@@ -0,0 +1,84 @@
+const { startClient, startNode, startTracker } = require('../../src/composition')
+const { BOOTNODES } = require('../../src/util')
+const { waitForEvent, LOCALHOST, DEFAULT_TIMEOUT } = require('../util')
+const TrackerNode = require('../../src/protocol/TrackerNode')
+const TrackerServer = require('../../src/protocol/TrackerServer')
+const NodeToNode = require('../../src/protocol/NodeToNode')
+
+jest.setTimeout(DEFAULT_TIMEOUT)
+
+describe('tracker, two nodes, publisher and two subscribers', () => {
```
harbu: Maybe have a general description where of what is being tested.

test/integration/subscriber.test.js
```diff
@@ -0,0 +1,84 @@
...
+        const publisherInterval = setInterval(() => {
+            publisher.publish(streamId, 'Hello world, from Publisher ' + publisher.protocols.nodeToNode.getAddress(), () => {})
+        }, 1000)
+
+        await waitForEvent(nodeOne.protocols.nodeToNode, NodeToNode.events.DATA_RECEIVED)
+        await waitForEvent(tracker.protocols.trackerServer, TrackerServer.events.STREAM_INFO_REQUESTED)
+        await waitForEvent(nodeOne.protocols.trackerNode, TrackerNode.events.STREAM_ASSIGNED)
+
+        const subscribeInterval1 = setInterval(() => {
+            subscriber1.subscribe(streamId)
```
harbu, line 54: Shouldn't one subscribe be enough?

harbu (changes requested):
Looking good. See comments.

src/protocol/TrackerNode.js
harbu: It is expected to get because common connection with NodeToNode and TrackerNode.

PR #100, opened 2018-09-26
Allow node to subscribe and fix message buffering test

harbu:
- Introduced two new classes that are used by `logic/Node`.
  - `SubscriberManager` is responsible for keeping track of who is subscribed to this node. It fires callbacks when a) first subscriber to a stream appears, and b) last one leaves. Most of the existing subscriber logic is moved to this class.
  - `SubscriptionManager` is responsible for keeping track of this node's subscriptions to other nodes.
- Allow for pending subscriptions via `SubscriptionManager`. If either tracker is not available or stream is still unknown, subscription cannot happen. We save these as pending and try later when stream info becomes available or tracker is connected to.
- `Node#onDataReceived` sends data to leader for (eventual) numbering. Added some TODOs for numbering branch.
- Allow Node to subscribe to a stream without client subscribers.

All this allowed for the fixing of test `nodeMessageBuffering.test.js`

TODO:
- Write unit tests for new classes

### DIFF
--- a/src/logic/Node.js
+++ b/src/logic/Node.js
@@ -2,6 +2,7 @@ const { EventEmitter } = require('events')
 const createDebug = require('debug')
 const TrackerNode = require('../protocol/TrackerNode')
 const NodeToNode = require('../protocol/NodeToNode')
+const SubscriberManager = require('../logic/SubscriberManager')
 const MessageBuffer = require('../helpers/MessageBuffer')
 const { getAddress, getIdShort } = require('../util')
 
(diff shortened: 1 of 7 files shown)


mirpo (approved):
maybe we need StreamsManager?
and don't forget to move tickets to "in progress" and assign ticket to yourself.

harbu:
Ah yeah forgot again... sorry about that.

Yeah I think we need that too. I will finish writing unit tests here first though.

Issue #95, opened 2018-09-25
Extract stream management logic to its own class


harbu:
Class `logic/Node` is becoming large with all the stream management (ownStreams, knownStreams, numbering, subscribers). Maybe we could extract this logic to a class (or several). The benefits would be:
- Separation of concerns
- Easier to unit test logic without having to start a libp2p node

PR #101, opened 2018-09-26
Encapsulate stream management logic to own class

harbu:
Fixes #95 

- Encapsulate stream management logic from `logic/Node` to its own class `StreamManager`.
- Use leader terminology


mirpo (approved):
coool

PR #105, opened 2018-09-27
try speedup travis

harbu:


harbu:
no effect + Kare tried earlier already

PR #110, opened 2018-09-28
Refactor string messages to objects

mirpo:


src/NetworkNode.js
```diff
@@ -8,11 +9,16 @@ module.exports = class NetworkNode extends Node {
...
+        const dataMessage = new DataMessage()
+        dataMessage.setStreamId(streamId)
+        dataMessage.setPayload(content)
+
+        this.onDataReceived(dataMessage)
     }
 
     addMessageListener(cb) {
-        this.on(Node.events.MESSAGE_RECEIVED, (streamId, content) => cb(streamId, 0, content))
+        this.node.on(Node.events.MESSAGE_RECEIVED, (dataMessage) => cb(dataMessage.getStreamId(), 0, dataMessage.getPayload()))
```
harbu: `this.node.on` should be `this.on`

src/helpers/MessageEncoder.js
```diff
@@ -22,15 +27,36 @@ const encode = (type, data) => {
...
+                version, code, source, data
+            })
+        default:
+            return Object.assign(new BasicMessage(), {
+                version, code, source, data
+            })
     }
 }
 
+// const createMessage = (code) => new BasicMessage(CURRENT_VERSION, code)
```
harbu: Just remove commented out code?

src/logic/Node.js
```diff
@@ -41,17 +42,10 @@ class Node extends EventEmitter {
...
-        })
-        this.protocols.nodeToNode.on(NodeToNode.events.DATA_RECEIVED, ({ streamId, data }) => this.onDataReceived(streamId, data))
-        this.protocols.nodeToNode.on(NodeToNode.events.SUBSCRIBE_REQUEST, ({ streamId, sender }) => {
-            this.onSubscribeRequest(streamId, sender)
-        })
-
-        this.protocols.nodeToNode.on(NodeToNode.events.UNSUBSCRIBE_REQUEST, ({ streamId, sender }) => {
-            this.onUnsubscribeRequest(streamId, sender)
-        })
+        this.protocols.trackerNode.on(TrackerNode.events.STREAM_INFO_RECEIVED, (streamMessage) => this.addKnownStreams(streamMessage))
```
harbu, line 45: 👍

harbu: Good that parameter names match class types, makes code more readable.

src/helpers/MessageEncoder.js
```diff
@@ -22,15 +27,36 @@ const encode = (type, data) => {
...
+        case msgTypes.STREAM:
+            return Object.assign(new StreamMessage(), {
+                version, code, source, data
+            })
+        case msgTypes.DATA:
+            return Object.assign(new DataMessage(), {
+                version, code, source, data
+            })
+        case msgTypes.SUBSCRIBE:
+        case msgTypes.UNSUBSCRIBE:
```
harbu, line 47: Do you think we need separate `UnsubscribeMessage` in future?

src/helpers/MessageEncoder.js
```diff
@@ -22,15 +27,36 @@ const encode = (type, data) => {
...
+const decode = (source, message) => {
     const { version, code, data } = JSON.parse(message)
-    return {
-        version,
-        code,
-        data
+
+    switch (code) {
+        case msgTypes.STATUS:
+            return Object.assign(new StatusMessage(), {
```
harbu, line 35: Why not constructor arguments?

src/logic/Node.js
```diff
@@ -185,7 +185,14 @@ class Node extends EventEmitter {
 
     _handleBufferedMessages(streamId) {
         this.messageBuffer.popAll(streamId)
-            .forEach((data) => this.onDataReceived(streamId, data))
+            .forEach((payload) => {
+                const dataMessage = new DataMessage()
+                dataMessage.setStreamId(streamId)
+                dataMessage.setPayload(payload)
+
+                // bad idea to call events directly
```
harbu: Could you put `TODO:` in front? Makes easier to `grep` these sorts of things later on. Regarding invoking event handler directly, I guess we could wrap the thing we really want to do into a separate function that both the event handler and this function invokes.

src/messages/DataMessage.js
```diff
@@ -0,0 +1,19 @@
...
+module.exports = class DataMessage extends BasicMessage {
+    getStreamId() {
+        return this.data[0]
+    }
+
+    setStreamId(streamId) {
+        this.data[0] = streamId
+    }
+
+    getPayload() {
```
harbu, line 12: I like this naming. `Data` for the protocol message stuff and `payload` for the actual content in the message. Let's try to use this separation in all code. 👍

src/messages/SubscribeMessage.js
```diff
@@ -0,0 +1,19 @@
...
+module.exports = class StreamMessage extends BasicMessage {
+    getStreamId() {
+        return this.data
+    }
+
+    setStreamId(streamId) {
+        this.data = streamId
+    }
+
+    getSender() {
```
harbu, line 12: Should just use `getSource()` as elsewhere?

src/connection/Libp2pEndpoint.js
```diff
@@ -64,10 +65,7 @@ class Libp2pEndpoint extends EventEmitter {
                 pull.drain((message) => {
                     debug('received from peer %s message with data "%s"', sender instanceof PeerInfo ? getAddress(sender) : '', message)
 
-                    this.emit(events.MESSAGE_RECEIVED, {
-                        sender,
-                        message
-                    })
+                    this.emit(events.MESSAGE_RECEIVED, encoder.decode(sender, message))
```
ptesavol: Here message is being decoded on the connection layer. Messages should be decoded on the protocol layer (what if there is a protocol that wants the raw data? we can never know)

src/helpers/MessageEncoder.js
ptesavol, line 35: Wouldn't it be better to use constructors instead?

src/logic/Node.js
```diff
@@ -185,7 +185,14 @@ class Node extends EventEmitter {
 
     _handleBufferedMessages(streamId) {
         this.messageBuffer.popAll(streamId)
-            .forEach((data) => this.onDataReceived(streamId, data))
+            .forEach((payload) => {
+                const dataMessage = new DataMessage()
+                dataMessage.setStreamId(streamId)
```
ptesavol, line 190: Would be better to use constructor parameters

src/protocol/TrackerServer.js
```diff
@@ -40,33 +40,24 @@ class TrackerServer extends EventEmitter {
...
         if (!isTracker(peer)) {
             this.emit(events.NODE_CONNECTED, peer)
         }
     }
 
-    onMessageReceived(peer, message) {
-        const { code, data } = encoder.decode(message)
-
-        switch (code) {
+    onMessageReceived(message) {
```
ptesavol, line 49: onMessageReceived() should receive raw data and do the decoding

ptesavol (changes requested):
Protocol layer should receive raw data, because we cannot make assumptions about the format of data a protocol wants to receive from the network. Message objects should be initialized using constructors instead of the ways that are used now. (there is also 4 comments in the code, most of them concern these 2 matters=

test/integration/libp2pendpoint.test.js
```diff
@@ -1,42 +1,46 @@
...
+
+        for (let i = 0; i < MAX; i++) {
+            expect(nodes[i].protocols.nodeToNode.endpoint.getPeers().length).toEqual(2)
+        }
 
         promises = []
         for (let i = 0; i < MAX; i++) {
-            // eslint-disable-next-line no-await-in-loop
-            promises.push(await endpoints[i].stop(console.log(`closing ${i} endpoint`)))
+            promises.push(nodes[i].protocols.nodeToNode.endpoint.stop(() => console.log(`closing ${i} endpoint`)))
```
harbu: should we use `node.stop()`?

test/integration/publisher.test.js
```diff
@@ -1,39 +0,0 @@
-const assert = require('assert')
```
harbu, line 1: This test was deleted. No longer needed?

harbu (commented):
Looks good. Few questsions; See comments

Issue #118, opened 2018-10-01
Protocol layer dependency on Libp2pEndpoint


ptesavol:
Protocol layer now depends on a specific Endpoint implementation (Libp2pEndpoint). This dependency should be removed to make it possible to write other Endpoints in the future ( WebsocketEndpoint, for example)

PR #119, opened 2018-10-01
Removed protocol layer dependency on Libp2pEndpoint

ptesavol:
Moved event definitions from Libp2pEndpoint to Endpoint interface class that Libp2pEndpoint implements. Now protocol layer is not dependent on a specific Endpoint type, and for example, implementing a WebsocketEndpoint is possible. Fix Issue #118.

### DIFF
--- a/src/protocol/TrackerServer.js
+++ b/src/protocol/TrackerServer.js
@@ -1,7 +1,6 @@
 const { EventEmitter } = require('events')
 const { isTracker, getAddress } = require('../util')
 const encoder = require('../helpers/MessageEncoder')
-const endpointEvents = require('../connection/Libp2pEndpoint').events
 const EndpointListener = require('./EndpointListener')
 
 const events = Object.freeze({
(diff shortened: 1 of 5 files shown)


src/protocol/TrackerServer.js
```diff
@@ -20,8 +19,6 @@ class TrackerServer extends EventEmitter {
 
         this._endpointListener = new EndpointListener()
         this._endpointListener.implement(this, endpoint)
-
-        this.endpoint.on(endpointEvents.PEER_DISCONNECTED, (peer) => this.onPeerDisconnected(peer))
```
harbu, line 24: Did you get rid of this on purpose? The callback is still needed.

harbu (changes requested):
LGTM, but one thing is missing imo. See comment.

src/protocol/TrackerServer.js
ptesavol: Yes, I removed it on purpose because the listener is already set in EndpointListener. This line I removed was a duplicate left from old code before the implementation of EndpointListener.

harbu: Ah I see now.

harbu (approved):

mirpo (approved):

Issue #111, opened 2018-10-01
Create graph network structure


harbu:
In the current version, a node that wants to subscribe to a stream will always subscribe to the leader of that stream. This creates a simplified topology, where messages are only passed from leader to subscriber.
 
We need the leader to be able to re-direct the node to subscribe to another subscriber.

ptesavol:
I think the tracker should create the random graph topology for stream S in the following way:

Every time a peer requests peers for Stream S from the tracker of stream S, do:
- Place all peers of stream S in an array (including the leader)
- Take K peers at random from the array and return to the requesting peer

harbu:
Upon further digging into code I agree. Do you however think we should somehow also return the leader if the requesting peer is interested in producing? I was thinking of responding with a random subset of subscribed peers as well as the leader?

ptesavol:
I don't know..returning the leader separately makes it very easy to implement a hacked client that always connects to the leader, and thus gets a lower latency. In the other hand, is there any way to prevent this kind of behaviour? Could we somehow check that a node is connected to exactly those nodes it got from the tracker and to nobody else?

PR #120, opened 2018-10-02
Repeater nodes and DAG subscription graph

harbu:
Added concept of _repeater_ node to code. Nodes now subscribe to repeaters for the data of a stream. The subscriber network still doesn't exhibit cycles, but rather forms a DAG. This is because a new subscriber will always have inbound edges formed only. To get cycles we need the ability to unsubscribe from streams.

Changes:
-  Add the subscriptions a node has to its status message. This allows tracker to know which streams a node is repeating.
- Send status message to tracker when a new subscription is created. This keeps tracker up-to-date with streams that node is repeater for.
- Include a random subset of repeaters in `streamMessage` tracker response to node. Node stores this information into `StreamManager`.
- Also keep leader in `streamMessage` in case producing to stream is needed on Node.
- Node subscribes to repeaters instead of leader directly to get data for stream.

Fixes #111


ptesavol:
Isn't it so that every subscriber is automatically a repeater?

harbu:
Good point. I hadn't thought about the distinction. 

You are correct that in this implementation every subscriber is a repeater. Unless the subscriber is a client, who does not participate in the peer-to-peer network. (Subscriber is now a bit overloaded because both clients and nodes may be subscribers.)

If we wanted to change it so that the subscriber gets to decide which streams it wishes to become a repeater of, it could send to the tracker in the status message only those subscriptions. In this case a better name for the statusMessage key `subscribedToStreams ` would be `repeaterOfStreams`. Current naming hints more towards the logic in which the tracker makes the decision based on information gathered from nodes.

ptesavol:
I think it makes no sense to let the nodes decide which streams to repeat and which not. The scalability of the network depends on the fact that those nodes who consume bandwidth also contribute bandwidth. On the contrary, in the future we need to find the best way to enforce that this happens, and discourage free-riding. There are many known methods for doing this, and it is only matter of experimenting to find the one that suits us best.

mirpo (approved):

PR #128, opened 2018-10-05
Duplicate message detection and avoidance

harbu:
- Class `DuplicateMessageDetector` that detects duplicate message even when they are arrive out-of-order and number gaps form.
- Instantiate and store detectors for each stream in `StreamManager`.
- In class `Node`:
   - Do not forward to subscribers or emit event when received message is deemed a duplicate.
   - Record metrics of duplicate messages
- Integration test that verifies that message duplication detection and avoidance works.


ptesavol (approved):
Looks ok.

Issue #114, opened 2018-10-01
Multiple trackers


harbu:
After #98 (DHT) look at multiple trackers

harbu:
Done.

Issue #113, opened 2018-10-01
Get rid of BOOTNODES global


harbu:
The main point is to make `isTracker()` work on non-global variable

harbu:
Done.

PR #132, opened 2018-10-17
WS + N trackers + remove global BOOTNODES

### LINKED ISSUE #125: Remove or fix test/integration/tracker.test.js
Currently commented out. Should be fixed or removed.

mirpo:
- Add support of Multiple trackers
- Add setBootnodes function for the Node
- Remove PEER_DISCOVERED event
- Remove libp2p files add WS

fixes #125, #122, #114, #113


.idea/jsLibraryMappings.xml
```diff
@@ -2,6 +2,5 @@
 <project version="4">
   <component name="JavaScriptLibraryMappings">
     <file url="file://$PROJECT_DIR$" libraries="{Node.js Core}" />
-    <file url="PROJECT" libraries="{@types/jest}" />
```
harbu: Let's keep this in. Allows IntelliJ to know that tests are Jest.

publisher.js
```diff
@@ -1,20 +1,19 @@
...
 
 const port = process.argv[2] || 30301
 const nodeAddress = process.argv[3] || ''
 const streamId = process.argv[4] || ''
 
-createEndpoint('127.0.0.1', port, '', true).then((endpoint) => {
-    endpoint.connect(nodeAddress)
+startClient('127.0.0.1', port, 'publisher1', nodeAddress)
+    .then((client) => {
+        client.protocols.nodeToNode.endpoint.connect(nodeAddress)
```
harbu, line 9: Hmm.. shouldn't the client do this internally?

src/composition.js
```diff
@@ -1,38 +1,63 @@
...
 const NodeToNode = require('./protocol/NodeToNode')
 const Tracker = require('./logic/Tracker')
 const Node = require('./logic/Node')
 const Client = require('./logic/Client')
 const NetworkNode = require('./NetworkNode')
+const WsEndpoint = require('./connection/WsEndpoint')
 
-async function startTracker(host, port, privateKey) {
-    return createEndpoint(host, port, privateKey, false).then((endpoint) => {
+async function WsNode(host, port) {
```
harbu: Talked about moving this to connection layer in meeting.

src/composition.js
```diff
@@ -1,38 +1,63 @@
...
+        })
+    })
+}
+
+async function createEndpoint(host, port, id) {
+    return WsNode(host, port).then((n) => new WsEndpoint(n, id))
+}
+
+async function startTracker(host, port, id) {
+    return createEndpoint(host, port, id, false).then((endpoint) => {
```
harbu: Can remove last boolean argument from `createEndpoint`.

src/composition.js
```diff
@@ -1,38 +1,63 @@
...
         return new Tracker(new TrackerServer(endpoint))
     }).catch((err) => {
         throw err
     })
 }
 
-async function startNode(host, port, privateKey, bootstrapTrackers) {
-    return createEndpoint(host, port, privateKey, true, bootstrapTrackers).then((endpoint) => {
+async function startNode(host, port, id) {
+    return createEndpoint(host, port, id, true).then((endpoint) => {
```
harbu: Can remove last boolean argument from `createEndpoint`.

src/connection/WsEndpoint.js
```diff
@@ -0,0 +1,121 @@
...
+        this.wss = node
+        this.id = id || uuidv4()
+
+        this.endpoint = new Endpoint()
+        this.endpoint.implement(this)
+
+        this.connections = new Map()
+
+        this.wss.on('connection', (ws, req) => {
+            const parameters = url.parse(req.url, true)
```
harbu, line 20: We might want to add validation here that parameters were indeed given.

src/connection/WsEndpoint.js
```diff
@@ -0,0 +1,121 @@
...
+            ws = this.connections.get(recipient)
+        } else {
+            debug('trying to send not existing socket %s', recipient)
+            return false
+        }
+
+        try {
+            ws.send(message)
+        } catch (e) {
+            console.error('Sorry, the web socket at "%s" is un-available')
```
harbu: %s, not assigned to.

harbu: Also return false here

src/connection/WsEndpoint.js
```diff
@@ -0,0 +1,121 @@
...
+
+        return true
+    }
+
+    isConnected(socketAddress) {
+        return this.connections.has(socketAddress)
+    }
+
+    async onReceive(sender, message) {
+        // debug('received from peer %s message with data "%s"', sender, message)
```
harbu: Lets keep this?

src/connection/WsEndpoint.js
```diff
@@ -0,0 +1,121 @@
...
+
+        this.emit(Endpoint.events.MESSAGE_RECEIVED, {
+            sender,
+            message
+        })
+    }
+
+    connect(peerAddress) {
+        if (this.connections.has(peerAddress)) {
+            return
```
harbu: Maybe we could add `debug` here to see if this happens a lot.

src/connection/WsEndpoint.js
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        return true
+    }
+
+    isConnected(socketAddress) {
+        return this.connections.has(socketAddress)
+    }
+
+    async onReceive(sender, message) {
```
harbu: Can remove `async`

src/connection/WsEndpoint.js
```diff
@@ -0,0 +1,121 @@
...
+        // debug('received from peer %s message with data "%s"', sender, message)
+
+        this.emit(Endpoint.events.MESSAGE_RECEIVED, {
+            sender,
+            message
+        })
+    }
+
+    connect(peerAddress) {
```
harbu, line 71: We could make this return a Promise so that user's can wait for connection be done (or refused). Maybe make testing easier?

src/connection/WsEndpoint.js
```diff
@@ -0,0 +1,121 @@
...
+                debug('failed to connect to %s, error: %o', peerAddress, err)
+            })
+        } catch (err) {
+            debug('failed to connect to %s, error: %o', peerAddress, err)
+        }
+    }
+
+    _onConnected(ws, peerId) {
+        // eslint-disable-next-line no-param-reassign
+        ws.peerId = peerId
```
harbu: Let's refactor later to  `Peer` (or `Connection`?) object

src/connection/WsEndpoint.js
```diff
@@ -0,0 +1,121 @@
...
+        this.connections.set(peerId, ws)
+
+        debug('connected -> %s', ws.peerId)
+        this.emit(Endpoint.events.PEER_CONNECTED, ws.peerId)
```
harbu: Let us move this emit to be last line of this function.

src/logic/Client.js
```diff
@@ -6,14 +6,18 @@ module.exports = class Client extends EventEmitter {
     constructor(nodeToNode, nodeAddress) {
         super()
 
-        this.id = getIdShort(nodeToNode.endpoint.node.peerInfo) // TODO: better way?
+        this.id = getIdShort(nodeToNode.endpoint.id)
```
harbu: I think we need better way still. Maybe pass id in constructor or something. Dunno.

src/logic/Node.js
```diff
@@ -207,24 +217,36 @@ class Node extends EventEmitter {
...
-        if (this.tracker) {
-            this.protocols.trackerNode.sendStatus(this.tracker, this._getStatus())
-            this.debug('status sent to tracker %s', getIdShort(this.tracker))
+    _sendStatusToAllTrackers() {
+        this.trackers.forEach((tracker, _) => this._sendStatus(tracker))
+    }
+
+    _sendStatus(tracker) {
+        this.debug('sending status to tracker %s', getIdShort(tracker))
+        if (tracker) {
```
harbu: Think we can get rid of this check now and assume that `tracker` is always given here.

src/logic/Node.js
```diff
@@ -247,6 +269,54 @@ class Node extends EventEmitter {
...
     }
+
+    _clearPeerRequestInterval() {
+        if (this.peersInterval) {
+            clearInterval(this.peersInterval)
+            this.peersInterval = null
+        }
+    }
+
+    setBootnodes(bootstrapNodes, delay = 3000) {
```
harbu: We should probably rename `setBootTrackers` and first argument to `bootstrapTrackers` for now because that is our current assumption in the code.

Later we can rename to `setBootPeers` and `bootstrapPeers` when that is supported.

src/logic/Node.js
```diff
@@ -247,6 +269,54 @@ class Node extends EventEmitter {
...
+        // TODO validate ws path
+        this.bootstrapNodes = bootstrapNodes
+
+        this._peerDiscoveryTimer = setInterval(() => {
```
harbu: Let's make this execute immediately and then wait for `delay` to make it faster. https://stackoverflow.com/questions/6685396/execute-the-setinterval-function-without-delay-the-first-time

harbu: Rename to `trackerDiscoveryInterval`.

src/logic/Node.js
```diff
@@ -247,6 +269,54 @@ class Node extends EventEmitter {
...
+                this.protocols.trackerNode.sendPeerMessage(tracker)
+            }, 5000)
+        }
+    }
+
+    _getTracker() {
+        return this.trackers.get([...this.trackers.keys()][Math.floor(Math.random() * this.trackers.size)])
+    }
+
+    _isTracker(tracker) {
```
harbu, line 314: I think these need to be moved to protocol layer later on. We can't just rely on bootstrap list for checking if a peer is a tracker or a node (or both?)

src/protocol/TrackerNode.js
```diff
@@ -17,40 +17,20 @@ class TrackerNode extends EventEmitter {
...
-        this.peersInterval = null
     }
 
     sendStatus(tracker, status) {
         this.endpoint.send(tracker, encoder.statusMessage(status))
     }
 
-    requestStreamInfo(tracker, streamId) {
-        this.endpoint.send(tracker, encoder.streamMessage(streamId, ''))
+    sendPeerMessage(tracker) {
```
harbu: what about `requestPeers`

src/util.js
```diff
@@ -12,35 +8,15 @@ const callbackToPromise = (method, ...args) => {
...
-    }
-    throw new Error('Expected instance of PeerInfo, got ' + peerInfo)
-}
-
-const getIdShort = (input) => (input instanceof PeerInfo ? getId(input) : input).slice(-4)
-
-const generateClientId = (suffix) => `${suffix}/v${version}/${os.platform()}-${os.arch()}/nodejs`
-
-const isTracker = (tracker) => BOOTNODES.includes(tracker)
+// TODO remove both
```
harbu, line 11: Will do

subscriber.js
```diff
@@ -1,4 +1,4 @@
-const { createEndpoint } = require('./src/connection/Libp2pEndpoint')
+const { createEndpoint } = require('./src/composition')
```
harbu: Could just use `createClient` or `startClient` ?

test/integration/ws-endpoint.test.js
```diff
@@ -17,33 +17,43 @@ describe('create two endpoints and init connection between them', () => {
...
-        await wait(1000)
-
-        for (let i = 0; i < MAX; i++) {
-            expect(endpoints[i].getPeers().length).toEqual(2)
+            // eslint-disable-next-line no-await-in-loop
+            promises.push(await waitForEvent(endpoints[i], endpointEvents.PEER_CONNECTED))
         }
 
-        promises = []
+        Promise.all(promises)
```
harbu: You are not awaiting for them? Needs to be fixed I think.

harbu (changes requested):

.idea/jsLibraryMappings.xml
mirpo: ok, maybe auto refactoring broke it

publisher.js
mirpo: yep, but it was before that PR

src/connection/WsEndpoint.js
mirpo: nope :)

src/connection/WsEndpoint.js
mirpo: agree

src/connection/WsEndpoint.js
mirpo: I already created empty Peer object

src/protocol/TrackerNode.js
mirpo: true

test/integration/ws-endpoint.test.js
mirpo: hmmm, I'm pushing to array above

src/util.js
harbu: Will remove in a later PR. Leave as is for now.

harbu (approved):

PR #152, opened 2018-11-21
Remove PEERS message type

harbu:
When a node connects to a tracker, it requests a random list of nodes and then proceeds to connect to those. This is no longer needed as a node will connect to those nodes that share responsibility for same streams. These peers are requested from the tracker via `STREAM_INFO` messages.

Removed the code related to fetching random peers from tracker after connecting.


ptesavol (approved):
Looks good. The removed feature had probably been forgotten there since the times we simply needed some peers to connect to in order to test connectivity.

src/messages/messageTypes.js
```diff
@@ -2,7 +2,6 @@ const CURRENT_VERSION = require('../../package.json').version
 
 const msgTypes = {
     STATUS: 0x00,
```
mirpo, line 4: Should numbering be updated? to be from 0 -> 5. for integrity

mirpo (approved):

PR #153, opened 2018-11-21
Use peer ids on logic layer

harbu:
Branch is based on #152 instead of master 

- Do some general refactoring on WsEndpoint (1st commit)
  - Remove unnecessary async keywords
  - Move private methods to last
  - Refactor debug messages
  - Rename peer book to connection list in connection layer
- Move PeerBook to protocol layer
- Use Peer Ids in logic layer instead of network addresses


ptesavol (approved):
Looking good.

mirpo (approved):

PR #154, opened 2018-11-22
Multiple nodes for stream

harbu:
Based on #153 

Allow multiple nodes to take responsibility for a stream. This is a huge overhaul of the code (especially `Node` class). I got rid of code no longer needed but also did some pretty heavy re-organizing. Might want to go over this face-to-face if wanted. 

## Key changes
- Nodes are no longer assigned to a stream by a tracker. Instead nodes assign themselves to streams and inform the tracker about this via an updated status message.
- `SubscribeMessage` has a new flag `leechOnly` which allows the recipient node to determine whether a subscribing node intends to propagate messages of the stream or whether it subscribed only to listen.
- Removed `SubscriberManager` and `SubscriptionManager`. Moved simple bookkeeping responsibility to `StreamManager`.
- Remove concept of _knownStream_, it is no longer needed. A node is only interested in streams it participates in.

## Smaller changes
- Got rid of unused event `NO_AVAILABLE_TRACKERS`
- More suitable data structure for `Node#trackers` (Map to Set)
- Peer types are checked in protocol layer instead of logic layer.
- `setBootstrapTrackers(addressList)` replaced with `async addBootstrapTracker(address)` which can be `await`(ed) in tests.
- Put `DataMessage` directly in MessageBuffer (instead of constructing/destructing into object)
- Remove unnecessary `await` events from tests (they work without them)
- Rename `ownStreams` to `streams`

## TODO:
- Rename Node functions
- Write test for PeerBook


src/helpers/MessageEncoder.js
```diff
@@ -31,7 +31,7 @@ const decode = (source, message) => {
             return new DataMessage(payload.streamId, payload.data, payload.number, payload.previousNumber, source)
 
         case msgTypes.SUBSCRIBE:
-            return new SubscribeMessage(payload, source)
+            return new SubscribeMessage(payload.streamId, payload.leechOnly, source)
```
ptesavol, line 34: Can you elaborate on the meaning of leechOnly? Isn't leeching something that we want to prevent from happening in the network? Allowing leeching to happen will ruin the scalability of the network.

If we consider the so called "light nodes", in my opinion they are the same thing as "clients", no need to create a distinct category for them.

ptesavol (commented):
In my opinion, speaking about nodes "being responsible of a stream" is no longer correct, as all nodes who participate in a stream are equal. Maybe this "being responsible of a stream" is a concept that comes from the older designs of the network where the nodes were not equal.

I would suggest using standard BitTorrent terms instead, "nodes who have joined the stream" or "nodes who participate in the stream".

src/helpers/MessageEncoder.js
harbu: Leeching is something  basically a 'light node' or a client would do, they want to receive messages but are not propagating them. In the current code, the same `SubscribeMessage` is sent by both client and node to subscribe to another node. The fleg `leechOnly` is there to let the node know whether the connecting peer intends to propagate or not, i.e., should we add it to our list of inbound nodes or is it just outbound node.

src/logic/Node.js
```diff
@@ -62,135 +53,100 @@ class Node extends EventEmitter {
...
-            this.protocols.trackerNode.requestStreamInfo(tracker, streamId)
+            if (!this._isSubscribed(streamId)) {
+                this.subscribeToStream(streamId)
+            }
         }
     }
 
-    _sendToSubscribers(dataMessage) {
+    _isReady(streamId) {
+        return this.streams.getOutboundNodesForStream(streamId).length > 0
```
harbu, line 100: Require at least one outbound node (subscriber) before propagating messages.

src/helpers/MessageEncoder.js
ptesavol: I would remove this option completely, and start thinking about ways to discourage leeching in general.

harbu: Think of it as a temporary thing for now. Let's remove this later when we have a better solution and time to solve it. Right now for the network to function properly we need a mechanism to differentiate between subscribing clients and nodes.

harbu: Regarding the leeching problem in the "final" network or later version of it: having the data  end-to-end encrypted and paying nodes only when they deliver data will probably discourage leeching quite well?

ptesavol: It is ok to leave this for now. Is it so that we kind of emulate a client with the node code using this flag? Don't we have the real client code available as well that we can use? 

In the final version, it saves a lot of upload bandwidth of a node if it just leeches. So having this option easily available is not what we want, because it is very tempting for the nodes to use it.

harbu: Yeah if you look at `logic/Client.js` it sends a subscribe message to a node via `NodeToNode` protocol. Problem was that node was counting this as an inbound node even though it only acts as a output node (because it only leeches)

mirpo (approved):

PR #164, opened 2018-12-18
Storage node (WIP)

harbu:
Add `startFullNode` function to composition.js which will start up a node that stores incoming `DataMessage` into Cassandra using wrapper class `Storage`.

TODO: Need to set up Cassandra onto Travis


harbu:
Closing this as we decided to add this functionality to Broker https://github.com/streamr-dev/broker instead of here.

PR #167, opened 2019-01-15
Move event MESSAGE_RECEIVED to correct place, add new MESSAGE_PROPAGATED event

mirpo:


bin/.idea/workspace.xml
```diff
@@ -0,0 +1,277 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
harbu: This file should not be commited. Maybe add into .gitignore?

harbu (changes requested):
LGTM; the workspace.xml needs to be removed.

PR #168, opened 2019-01-16
Add more error handling for connections

mirpo:
- fix node errors when it starts without tracker
- publisher was dying when node leaves
- publisher tries to reconnect if desired node is away

### DIFF
--- a/src/logic/Node.js
+++ b/src/logic/Node.js
@@ -172,7 +172,12 @@ class Node extends EventEmitter {
 
     _requestStreamInfo(streamId) {
         const randomTracker = this._getTracker()
-        this.protocols.trackerNode.requestStreamInfo(randomTracker, streamId)
+
+        if (randomTracker) {
+            this.protocols.trackerNode.requestStreamInfo(randomTracker, streamId)
+        } else {
+            console.error('Not connected to any tracker')
+        }
     }
 
     async _subscribeToStreamOnNode(node, streamId) {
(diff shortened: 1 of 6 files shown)


src/logic/Node.js
```diff
@@ -203,7 +208,9 @@ class Node extends EventEmitter {
 
     async addBootstrapTracker(trackerAddress) {
         this.bootstrapTrackerAddresses.push(trackerAddress)
-        await this.protocols.trackerNode.connectToTracker(trackerAddress)
+        await this.protocols.trackerNode.connectToTracker(trackerAddress).catch((err) => {
```
harbu: I think we should change this to
```
return this.protocols.trackerNode.connectToTracker(trackerAddress)
```
then make it the responsibility of the user of this class to handle the error.

E.g.
```
node.addBootstrapTracker('ws://127.0.0.1:8888')
  .catch((e) => console.error(`Could not connect to tracker ${trackerAddress} because '${err}'`)
```

This allows user of class to decide how to handle error. Maybe logging is enough, but maybe sometimes the whole program should be restarted etc. etc.

What do you think?

src/protocol/NodeToNode.js
```diff
@@ -29,7 +29,7 @@ class NodeToNode extends EventEmitter {
 
     sendData(receiverNodeId, messageId, previousMessageReference, payload) {
         const receiverNodeAddress = this.peerBook.getAddress(receiverNodeId)
-        this.endpoint.send(receiverNodeAddress, encoder.dataMessage(messageId, previousMessageReference, payload))
+        return this.endpoint.send(receiverNodeAddress, encoder.dataMessage(messageId, previousMessageReference, payload))
```
harbu: 👍

src/protocol/PeerBook.js
```diff
@@ -22,10 +22,11 @@ module.exports = class PeerBook {
     }
 
     getAddress(peerId) {
-        if (!this.hasPeerId(peerId)) {
-            throw new Error(`Id ${peerId} not found in peer book`)
+        if (this.hasPeerId(peerId)) {
```
harbu: Should we keep the old version since in `Client.js` you are already `catch`ing?

harbu (changes requested):

harbu (approved):

PR #169, opened 2019-01-18
Limit number of connections per stream

mirpo:
Added number of connections for the stream. But we have InboundNodesForStream and OutboundNodesForStream, so now I check sum of both. But maybe we need separate limits? 
And also now we can get disconnection reason on protocol level.
@harbu with MAX_NUM_OF_NODES_PER_STREAM = 6 message-duplication.test.js is failing. So I removed 5th node. But maybe we want to add more tests for some edge cases.

### DIFF
--- a/src/logic/Node.js
+++ b/src/logic/Node.js
@@ -131,47 +131,27 @@ class Node extends EventEmitter {
         const source = subscribeMessage.getSource()
         const leechOnly = subscribeMessage.getLeechOnly()
 
-        this.subscribeToStreamIfHaveNotYet(streamId)
+        const isSetup = this.streams.isSetUp(streamId)
 
-        if (this.streams.getOutboundNodesForStream(streamId).length < MAX_NUM_NODES_OUTBOUND_PER_STREAM) {
-            this.streams.addOutboundNode(streamId, source)
-        } else {
+        if (isSetup && this.streams.getOutboundNodesForStream(streamId).length >= MAX_NUM_NODES_OUTBOUND_PER_STREAM) {
             this.debug('reached max number "%d" for outbound connections for stream %s', MAX_NUM_NODES_OUTBOUND_PER_STREAM, streamId)
             this.protocols.nodeToNode.disconnectFromNode(source, disconnectionReasons.MAX_OUTBOUND_CONNECTIONS)
-        }
+        } else if (isSetup && !leechOnly && this.streams.getInboundNodesForStream(streamId).length >= MAX_NUM_NODES_INBOUND_PER_STREAM) {
+            this.debug('reached max number "%d" for inbound connections for stream %s', MAX_NUM_NODES_INBOUND_PER_STREAM, streamId)
+            this.protocols.nodeToNode.disconnectFromNode(source, disconnectionReasons.MAX_INBOUND_CONNECTIONS)
+        } else {
+            this.subscribeToStreamIfHaveNotYet(streamId)
 
-        if (!leechOnly) {
-            if (this.streams.getInboundNodesForStream(streamId).length < MAX_NUM_NODES_INBOUND_PER_STREAM) {
+            this.streams.addOutboundNode(streamId, source)
+
+            if (!leechOnly) {
                 this.streams.addInboundNode(streamId, source)
-            } else {
-                this.debug('reached max number "%d" for inbound connections for stream %s', MAX_NUM_NODES_INBOUND_PER_STREAM, streamId)
-                this.protocols.nodeToNode.disconnectFromNode(source, disconnectionReasons.MAX_INBOUND_CONNECTIONS)
             }
+
+            this._handleBufferedMessages(streamId)
+            this.debug('node %s subscribed to stream %s', source, streamId)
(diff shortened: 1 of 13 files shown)


src/protocol/NodeToNode.js
```diff
@@ -42,6 +42,13 @@ class NodeToNode extends EventEmitter {
         this.endpoint.send(receiverNodeAddress, encoder.unsubscribeMessage(streamId))
     }
 
+    sendDisconnectionMessage(receiverNodeId, reason) {
```
harbu: What about renaming to `disconnectFromNode` or `disconnect` since you are actually disconnecting (not just sending msg)?

src/logic/Node.js
```diff
@@ -130,13 +132,18 @@ class Node extends EventEmitter {
 
         this.subscribeToStreamIfHaveNotYet(streamId)
 
-        this.streams.addOutboundNode(streamId, source)
-        if (!leechOnly) {
-            this.streams.addInboundNode(streamId, source)
+        if (this.streams.getInboundNodesForStream(streamId).length + this.streams.getOutboundNodesForStream(streamId).length >= MAX_NUM_OF_NODES_PER_STREAM) {
```
harbu: Btw each connected node will have an entry in both `inboundNodes` and `outboundNodes`, so now the limit becomes `MAX_NUM_OF_NODES_PER_STREAM / 2` WebSocket connections to nodes (per stream)

test/integration/message-duplication.test.js
```diff
@@ -22,8 +22,7 @@ describe('duplicate message detection and avoidance', () => {
             startNetworkNode(LOCALHOST, 30352, 'node-1'),
             startNetworkNode(LOCALHOST, 30353, 'node-2'),
```
harbu: Hmm.. I think this test should still work though? Because it is only checking upper limits not exact numbers. I'll try it out before change and see if I can see what the issue is.

harbu (changes requested):
LGTM; few ideas + I think the test should still work so we should figure out what is the issue?

test/integration/message-duplication.test.js
mirpo: I put original test back, lets check together.

src/logic/Node.js
```diff
@@ -132,18 +133,25 @@ class Node extends EventEmitter {
...
+            this.debug('reached max number "%d" for outbound connections for stream %s', MAX_NUM_NODES_OUTBOUND_PER_STREAM, streamId)
+            this.protocols.nodeToNode.disconnectFromNode(source, disconnectionReasons.MAX_OUTBOUND_CONNECTIONS)
+        }
+
+        if (!leechOnly) {
+            if (this.streams.getInboundNodesForStream(streamId).length < MAX_NUM_NODES_INBOUND_PER_STREAM) {
                 this.streams.addInboundNode(streamId, source)
+            } else {
+                this.debug('reached max number "%d" for inbound connections for stream %s', MAX_NUM_NODES_INBOUND_PER_STREAM, streamId)
+                this.protocols.nodeToNode.disconnectFromNode(source, disconnectionReasons.MAX_INBOUND_CONNECTIONS)
```
harbu: Method `nodeToNode.disconnectFromNode` may be invoked the 2nd time here if the earlier `else` block got executed?

src/logic/Node.js
```diff
@@ -132,18 +133,25 @@ class Node extends EventEmitter {
...
             }
-            this._handleBufferedMessages(streamId)
-            this.debug('node %s subscribed to stream %s', source, streamId)
-            this.emit(events.SUBSCRIPTION_RECEIVED, streamId, source)
         }
+
+        this._handleBufferedMessages(streamId)
```
harbu: I think this code should get executed only if `disconnectFromNode` was not executed.

test/integration/connection-limits.test.js
```diff
@@ -0,0 +1,57 @@
+const { startNetworkNode, startTracker } = require('../../src/composition')
+const { callbackToPromise } = require('../../src/util')
+const { LOCALHOST, DEFAULT_TIMEOUT } = require('../util')
+const endpointEvents = require('../../src/connection/Endpoint').events
+const { disconnectionReasons } = require('../../src/messages/messageTypes')
+
+jest.setTimeout(DEFAULT_TIMEOUT)
+
+/**
+ * This test verifies that on receiving a duplicate message, it is not re-emitted to the node's subscribers.
```
harbu, line 10: This comment needs to be removed or re-written

test/integration/connection-limits.test.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+describe('duplicate message detection and avoidance', () => {
```
harbu, line 12: Better description?

harbu (changes requested):

src/logic/Node.js
mirpo: Yes, but we check in WsEndpoint.close " if (!this.isConnected(recipientAddress))

ptesavol (approved):
Looks ok. Does the test go through now?

PR #177, opened 2019-02-13
 Add instructions from tracker to node to reconnect

mirpo:
I added:
- Instructions to reconnect from tracker
- Function for node to reconnect to the new nodes from tracker
- Added some check for ws.OPEN, minor additions to disconnect functions and etc.
- Added new disconnection reasons
- Added test for reconnection instructions, but super simple only with two nodes.

TODO @harbu:
- Check tests, we have weird situation when PeerBook raises and Error, especially on shutting down. Please check tests with --detectOpenHandles
- I commented part of message-propagation.test.js


src/connection/WsEndpoint.js
```diff
@@ -75,14 +75,19 @@ class WsEndpoint extends EventEmitter {
...
+                            if (err) {
+                                reject(err)
+                            } else {
+                                debug('sent to %s message "%s"', recipientAddress, message)
+                                resolve()
+                            }
+                        })
+                    } else {
+                        debug('sent failed because readyState of socket is %d', ws.readyState)
+                        resolve()
```
harbu, line 89: Should this be `reject()` ?

src/connection/WsEndpoint.js
```diff
@@ -75,14 +75,19 @@ class WsEndpoint extends EventEmitter {
...
                     const ws = this.connections.get(recipientAddress)
-                    ws.send(message, (err) => {
-                        if (err) {
-                            reject(err)
-                        } else {
-                            debug('sent to %s message "%s"', recipientAddress, message)
-                            resolve()
-                        }
-                    })
+                    if (ws.readyState === ws.OPEN) {
```
harbu, line 78: So this check is here in case we have disconnected but the disconnection hasn't yet been finalized?

src/logic/Node.js
```diff
@@ -81,13 +81,25 @@ class Node extends EventEmitter {
         }
     }
 
-    onStreamInfoReceived(streamMessage) {
+    async onStreamInfoReceived(streamMessage) {
         const streamId = streamMessage.getStreamId()
         const nodeAddresses = streamMessage.getNodeAddresses()
+        const alreadyConnectedNode = [...this.streams.getAllOutboundNodes()]
```
harbu: `alreadyConnectedNodes`

src/connection/WsEndpoint.js
harbu: Maybe a code comment would be good here to make sure no one deletes it by accident etc.

test/integration/connection-limits.test.js
```diff
@@ -49,7 +49,7 @@ describe('check maxInBound and maxOutBound limits', () => {
             // update limits and connect again
             otherNodes.forEach((node) => {
                 const limits = node.getConnectionLimitsPerStream()
-                node.setConnectionLimitsPerStream(limits.maxInBound + 2, limits.maxOutBound + 2)
+                node.setConnectionLimitsPerStream(limits.maxInBound + 1, limits.maxOutBound + 1)
```
harbu, line 52: Why this change?

test/integration/tracker-node-reconnect-instructions.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * This test verifies that tracker can send instructions to node and node will connect and disconnect based on the instructions
+ */
+describe('Check tracker instructions to node', () => {
+    let tracker
+    let otherNodes
+    const streamId = 'stream-1'
+
+    it('init tracker and nodes, tracker receives stream info', async (done) => {
```
harbu, line 21: Maybe change to `beforeAll` since no assertions and this is assumed to be ran before tests?

harbu: You can remove `done()` from end of this function is you change this function to `async () => {`. It is already async so it will complete after last `await`.

test/integration/tracker-node-reconnect-instructions.test.js
```diff
@@ -0,0 +1,76 @@
...
+            startNetworkNode(LOCALHOST, 30952, 'node-1'),
+            startNetworkNode(LOCALHOST, 30953, 'node-2')
+        ])
+        await Promise.all(otherNodes.map((node) => node.addBootstrapTracker(tracker.getAddress())))
+        await Promise.all(otherNodes.map((node) => node.subscribe(streamId, 0)))
+        await Promise.all(otherNodes.map((node) => waitForEvent(node.protocols.trackerNode, TrackerNode.events.STREAM_INFO_RECEIVED)))
+        done()
+    })
+
+    it('tracker should receive statuses from both', async (done) => {
```
harbu, line 34: keyword `async` not needed

test/integration/tracker-node-reconnect-instructions.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+            if (receivedTotal === otherNodes.length) {
+                done()
+            }
+        })
+    })
+
+    it('tracker sends empty list of nodes, so node-one will disconnect from node two', async (done) => {
+        // eslint-disable-next-line no-underscore-dangle
+        otherNodes.map((node) => node._clearMaintainStreamsInterval())
```
harbu, line 47: There is some reliance on internals here which may cause this test to require lots of fixing in future as we change internals. I think it is OK for now but in future we might want to take a more black-box approach here or, alternatively, structure our code to allow for more fine-grain testing.

harbu (changes requested):
Looking good; I'll look into the tests next.

src/connection/WsEndpoint.js
mirpo: Ok

src/connection/WsEndpoint.js
mirpo: Not sure, it's not an error, we just can't send.

src/logic/Node.js
mirpo: ?

test/integration/connection-limits.test.js
mirpo: Originally was not correct, we have limit 6 in code, and just need in test +1 to connect to the 7th node

test/integration/tracker-node-reconnect-instructions.test.js
mirpo: Ok

test/integration/tracker-node-reconnect-instructions.test.js
mirpo: Ok

test/integration/tracker-node-reconnect-instructions.test.js
mirpo: I didn't test it, but maybe we could remove this line at all.

src/logic/Node.js
```diff
@@ -81,13 +81,25 @@ class Node extends EventEmitter {
...
+        const nodeIds = []
 
-        nodeAddresses.forEach(async (nodeAddress) => {
+        await Promise.all(nodeAddresses.map(async (nodeAddress) => {
```
ptesavol, line 90: I would consider switching the order of connecting and disconnecting nodes. Now there will momentarily be nodes who have more outgoing connections than what connection limits would allow.

src/logic/Tracker.js
```diff
@@ -1,7 +1,7 @@
 const { EventEmitter } = require('events')
 const createDebug = require('debug')
 const TrackerServer = require('../protocol/TrackerServer')
-const { getPeersTopology } = require('../helpers/TopologyStrategy')
+const { getPeersTopology, filterOutRandomPeer } = require('../helpers/TopologyStrategy')
```
ptesavol, line 4: I would bring these helper functions from TopologyStrategy.js back to the Tracker.js file. This way there would be a single file for those people to edit who are interested in implementing different kinds of Trackers. Another idea is that maybe the tracker to be used in an experiment could be chosen with a command line parameter so the different trackers could be easier to compare.

src/logic/Tracker.js
```diff
@@ -34,12 +34,27 @@ module.exports = class Tracker extends EventEmitter {
...
         const source = streamMessage.getSource()
 
         const nodesForStream = this.streamKeyToNodes.get(streamId.key()) || new Set()
         const selectedNodes = getPeersTopology([...nodesForStream], source)
+
+        await Promise.all(selectedNodes.map((node) => {
+            let { outboundNodes } = this.connectionsToNodes.get(node)
+            if (outboundNodes.length) {
+                this.debug('original connections %j', outboundNodes)
+                outboundNodes = filterOutRandomPeer(outboundNodes)
```
ptesavol: Here you instruct the node to disconnect one outgoing connection, but what happens to the other end? The node in the other end will also have one more connection slot available after the disconnection. What happens to this slot?

ptesavol (changes requested):

PR #180, opened 2019-02-22
Extract overlay topology to own class

harbu:
First of all this is not yet ready. I have to
- integrate this to Tracker.js
- test it out in practice
- a few edge cases need unit test still

Just made the PR to allow @ptesavol to go further with his work and comment on the logic here.

The idea here is that on receiving a node status tracker will invoke `update()` and on node disconnect `leave()`. After both cases the tracker should invoke `formInsructions(nodeId)` to get a list of `connect` and `disconnect` instructions that need to be sent to `nodeId` as well as other nodes.

### DIFF
--- a/src/logic/OverlayTopology.js
+++ b/src/logic/OverlayTopology.js
@@ -1,4 +1,4 @@
-//https://gist.github.com/guilhermepontes/17ae0cc71fa2b13ea8c20c94c5c35dc4
+// From: https://gist.github.com/guilhermepontes/17ae0cc71fa2b13ea8c20c94c5c35dc4
 const shuffleArray = (arr) => arr
     .map((a) => [Math.random(), a])
     .sort((a, b) => a[0] - b[0])
--- a/src/logic/OverlayTopology.js
+++ b/src/logic/OverlayTopology.js
@@ -60,7 +60,7 @@ module.exports = class OverlayTopology {
                 .filter(([n, neighbors]) => neighbors.size < this.maxNeighborsPerNode) // there are open slots
                 .filter(([n, neighbors]) => !neighbors.has(nodeId)) // nodeId is not already a neighbor
                 .filter(([n, _]) => n !== nodeId) // do not connect to self
-                .sort(([n1, neighbors1], [n2, neighbors2]) => neighbors1.size - neighbors2.size)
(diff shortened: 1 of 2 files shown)


mirpo (approved):

src/logic/OverlayTopology.js
```diff
@@ -0,0 +1,101 @@
...
+                .map(([n, _]) => n)
+
+            // TODO: if nodeId is already connected to one participant of a connection that will be disconnected, we will
+            //  not be freeing up 2 slots but instead only 1. Should this be taken into consideration?
+            const disconnectionTargets = this.shuffleArray(candidates).slice(0, Math.floor(missingNeighbors / 2))
+            instructions.push(...disconnectionTargets.map((n) => {
+                const r = this.random() < 0.5
+                const randomNeighbor = this.pickRandomElement([...this.nodes[n]])
+                return {
+                    action: 'disconnect',
```
harbu: Would probably be better to use terms join/leave or subscribe/unsubscribe instead of connecting and disconnecting since two nodes might share other streams as well. The ultimate decision as to whether to connect or disconnect should be that of the nodes.

ptesavol (changes requested):
It seems that in this code the tracker is assumed to give instructions to the nodes in an imperative way, ie. "disconnect from node 24234". In the current tracker.js and node implementations, however, it is assumed that the instructions from the tracker are declarative, ie. "you should have outgoing connections to nodes: [213113,23323,53533,343434]" Both the real node and my testbench assume this declarative style. Maybe this code should be updated to follow the declarative model?

harbu:
Either way is fine by me. Do you think we should be able to differentiate the active participant in these cases though, i.e., who should open the socket and who should close the socket? Or should both participants try to do it, with the one succeeding who gets to it first?

ptesavol:
I think it is better to use the declarative style because it is already implemented in the real node and the testbench. Also the active participant should be the node on the outgoing connections side. The declarations for incoming connections currently are for information only, but this feature could be used later to block incoming connections from random peers not approved by the tracker.

harbu:
Alrighty, I will look into turning this to declarative style. Hopefully done today.

harbu:
Instructions are now in declarative style

PR #191, opened 2019-03-05
Ovelay topology integration with network + other stuff

harbu:
There are multiple things included here (stuff sort of snowballed)

1. Replace STREAM_INFO with INSTRUCTION.
2. Remove stream connection limits
3. Fix bug in duplicate socket closing
4. Integrate Tracker with OverlayTopology
5. Made errors in `PeerBook` into their own class `NotFoundInPeerBookError`
6. Fix tests

Currently there is a bit of a hack included where Tracker has a "limit" on the amount of instructions it can send to nodes. Originally I thought this was due to an inherent instability in our tracker algorithm. Now after finding the bug (_3. Fix bug in duplicate socket closing_) I think that is causing the issue. Hence, I'm now looking into removing this "limit" hack.


src/connection/WsEndpoint.js
```diff
@@ -212,6 +212,10 @@ class WsEndpoint extends EventEmitter {
         })
 
         ws.on('close', (code, reason) => {
+            if (reason === 'streamr:endpoint:duplicate-connection') {
```
mirpo: Move to disconnectionReasons

src/logic/Tracker.js
```diff
@@ -1,22 +1,22 @@
 const { EventEmitter } = require('events')
 const createDebug = require('debug')
 const TrackerServer = require('../protocol/TrackerServer')
-const { getPeersTopology, filterOutRandomPeer } = require('../helpers/TopologyStrategy')
+const OverlayTopology = require('../logic/OverlayTopology')
+const { StreamID } = require('../identifiers')
+
+const NEIGHBORS_PER_NODE = 4
```
mirpo, line 7: We need some function to change it

test/integration/tracker-node-reconnect-instructions.test.js
```diff
@@ -2,7 +2,7 @@ const { startNetworkNode, startTracker } = require('../../src/composition')
 const { callbackToPromise } = require('../../src/util')
```
mirpo, line 2: Maybe we need more tests for instructions here? Form instructions, send them, check connections and etc

mirpo (changes requested):
Looks good  :+1:

src/logic/OverlayTopology.js
```diff
@@ -26,8 +26,10 @@ module.exports = class OverlayTopology {
     }
 
     leave(nodeId) {
-        this.nodes[nodeId].forEach((neighbor) => this.nodes[neighbor].delete(nodeId)) // assumption: neighbor already initialized in tracker
-        delete this.nodes[nodeId]
+        if (this.nodes[nodeId] != null) {
```
mirpo, line 29: Add this plz at the end (and remove module.exports in top)

if (typeof module !== 'undefined' && module.exports) {
    module.exports = OverlayTopology
}

so we'll be able to reuse in browser

src/connection/WsEndpoint.js
harbu: done

src/logic/Tracker.js
harbu: Let's add it when the need comes?

test/integration/tracker-node-reconnect-instructions.test.js
harbu: Not doing this now. I created issue for this https://github.com/streamr-dev/network/issues/192 when time.

src/logic/OverlayTopology.js
harbu: done

PR #197, opened 2019-03-08
OverlayTopology: discard unknown nodes

harbu:
@mirpo discovered in emulator that in certain situations tracker will receive information about nodes that haven't yet sent their status to it. This could happen, for example, if two nodes `N1`, `N2` connect to tracker `T` and subscribe to a common stream. Then the tracker `T` goes down for a moment and comes up. In this scenario tracker `T` will receive status of either `N1` or `N2` first with the other node being unknown to tracker. 

Wrote handling for this into `OverlayTopology`. Unknown nodes will simply be discard from neighbors list.

### DIFF
--- a/src/logic/OverlayTopology.js
+++ b/src/logic/OverlayTopology.js
@@ -18,8 +18,10 @@ class OverlayTopology {
     }
 
     update(nodeId, neighbors) {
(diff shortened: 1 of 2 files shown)


mirpo (approved):
looks good

PR #207, opened 2019-03-19
Upd conn

mirpo:
As a result got better stability even with 500+ nodes after removing first empty sendStatus. Two tests are failing, I know, just skipped for now. And I have few more abnormal points to debug and fix.
Bonus - now it's possible to debug whole network through the IntelliJ IDEA.

And it's even possible to remove sendStatusTimeout, but I didn't test it in emulator yet.


src/logic/Node.js
```diff
@@ -270,9 +270,8 @@ class Node extends EventEmitter {
             })
     }
 
-    async addBootstrapTracker(trackerAddress) {
+    addBootstrapTracker(trackerAddress) {
         this.bootstrapTrackerAddresses.push(trackerAddress)
-        return this.protocols.trackerNode.connectToTracker(trackerAddress)
```
harbu, line 275: Hmm.. why remove `return`? With `return` a user of this method can `await` for the connection to happen, e.g., in tests.

harbu (changes requested):
LGTM; see one question I have

src/logic/Node.js
mirpo: Because we have _connectToBootstrapTrackers, few lines below with correct error handling, so why do we need duplicate code and immediate connection to tracker? And we can always await event CONNECTED_TO_TRACKER.

harbu: Ah I didn't notice the immediate connection was removed, okay, it makes sense then.

harbu: I think in tests this causes problem because `_connectToBootstrapTrackers` will only be invoked every 5 seconds.

mirpo: Yes, I'm fixing that

PR #210, opened 2019-03-22
WsEndpoint: only drop incoming duplicate socket (work in progress)

harbu:
Noticed while doing https://github.com/streamr-dev/network/pull/209 that there is still problem with duplicate socket disconnection. This is another attempt of fixing this by making sure we only disconnect in incoming connection scenario (as opposed to both incoming and outgoing, as before)


harbu:
This does not fix the issue. Closing.

PR #212, opened 2019-03-26
WsEndpoint: keep track of pending connections

harbu:
`WsEndpoint#connect(peerAddress)` may be invoked several times in a short period of time with the same argument `peerAddress`. Until a socket had finally been opened, this would cause multiple sockets to be opened on `peerAddress`, which would fool the duplicate socket avoidance logic to drop the only socket between the two nodes.

Fixed issue so that pending connections are kept track of, thereby avoiding multiple sockets to be opened on the same peer. Also keep track of promises for pending connections and resolve/reject when appropriate.


mirpo (approved):
LGFM

PR #213, opened 2019-04-01
Fix ready state

mirpo:
I left some extra console.log so you could recheck me, by commenting out one line with                 this.connections.get(address).close() and run npm run network N

### DIFF
--- a/src/connection/WsEndpoint.js
+++ b/src/connection/WsEndpoint.js
@@ -75,7 +75,7 @@ class WsEndpoint extends EventEmitter {
                 console.log('-------------> killing dublicate socket')
                 console.log('-------------> already connected to %s, readyState %d', address, this.connections.get(address).readyState)
 
-                this.connections.get(address).close(1000, disconnectionReasons.DUPLICATE_SOCKET)
+                this.connections.get(address).close()
             } else {
                 console.log('new income connection ok')
             }
--- a/src/connection/WsEndpoint.js
+++ b/src/connection/WsEndpoint.js
@@ -67,12 +67,37 @@ class WsEndpoint extends EventEmitter {
 
         this.wss.on('connection', this._onIncomingConnection.bind(this))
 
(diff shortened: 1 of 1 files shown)


src/connection/WsEndpoint.js
```diff
@@ -67,12 +67,38 @@ class WsEndpoint extends EventEmitter {
 
         this.wss.on('connection', this._onIncomingConnection.bind(this))
 
+        this.wss.options.verifyClient = (info) => {
+            const parameters = url.parse(info.req.url, true)
+            const { address } = parameters.query
+
+            if (this.isConnected(address)) {
```
harbu, line 74: Interesting. Do I understand correctly that in this case you drop existing socket if a new socket is established between the same two peers?

src/connection/WsEndpoint.js
```diff
@@ -67,12 +67,38 @@ class WsEndpoint extends EventEmitter {
...
+            return true
+        }
+
         // Attach custom headers to headers before they are sent to client
         this.wss.on('headers', (headers) => {
             headers.push(...this.customHeaders.asArray())
         })
 
         debug('listening on: %s', this.getAddress())
+        this.checkConnections = setInterval(this._checkConnections.bind(this), 2000)
```
harbu: Good idea. Let's keep this to make sure that no problems with readyState in future?

src/connection/WsEndpoint.js
harbu: Do we still need the duplicate socket dropping in `this._onNewConnection`?

harbu (changes requested):
tests need to be fixed; LGTM otherwise.

src/connection/WsEndpoint.js
mirpo: Yes

src/connection/WsEndpoint.js
mirpo: :thumbsup:

PR #220, opened 2019-04-04
Resend message types & encoding/decoding

harbu:
Add 7 new message types for resends based on https://github.com/streamr-dev/streamr-client-protocol-js/blob/master/PROTOCOL.md and implement encoding/decoding of them.

### DIFF
--- a/src/messages/messageTypes.js
+++ b/src/messages/messageTypes.js
@@ -1,12 +1,19 @@
 const CURRENT_VERSION = require('../../package.json').version
 
 const msgTypes = {
-    STATUS: 0x00,
-    DATA: 0x02,
-    SUBSCRIBE: 0x03,
-    UNSUBSCRIBE: 0x04,
-    PUBLISH: 0x05,
-    INSTRUCTION: 0x06
+    STATUS: 0,
+    DATA: 1,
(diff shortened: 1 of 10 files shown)


src/messages/messageTypes.js
```diff
@@ -1,12 +1,19 @@
...
-    UNSUBSCRIBE: 0x04,
-    PUBLISH: 0x05,
-    INSTRUCTION: 0x06
+    STATUS: 0,
+    DATA: 1,
+    SUBSCRIBE: 2,
+    UNSUBSCRIBE: 3,
+    PUBLISH: 4,
+    INSTRUCTION: 5,
+    RESEND_LAST: 6,
```
mirpo, line 10: OMG......... Why not one RESEND with sub types.....

mirpo (approved):

src/messages/messageTypes.js
harbu: Protocol naming and structure is based on current Data API and [PROTOCOL.md](https://github.com/streamr-dev/streamr-client-protocol-js/blob/master/PROTOCOL.md). We should probably conform to that since the roadmap includes harmonizing the protocols into one.

Issue #222, opened 2019-04-05
Tracker may assign node to be its own neighbor


harbu:
In some situations a tracker may assign a node to be its own neighbor. The problem seems to be stemming from `OverlayTopology`.

It occurs once in a while when running test referenced #224

PR #225, opened 2019-04-06
OverlayTopology: fix issue where nodes get assigned as their own neighbor

harbu:
Fixes #222. 

Also added invariant check to end of `formInstructions` that can be maybe later deleted but is useful for now as verification that the state of the object is as expected (i.e. no node has itself as neighbor)


mirpo (approved):
Goooood, looks like it's better to rerun experiments with this fix?

PR #229, opened 2019-04-08
PeerBook: throw error if metadata not given

harbu:
Fail-fast if `streamr-peer-id` or `streamr-peer-type` not given in websocket headers.


mirpo (approved):
LGTM, but some problem with tests

harbu:
Seems to be flaky I wrote #230 issue about it.

PR #232, opened 2019-04-09
Storage node registration in tracker

mirpo:


src/logic/Node.js
```diff
@@ -93,6 +94,10 @@ class Node extends EventEmitter {
 
         this.debug('received instructions for %s', streamId)
 
+        if (this.protocols.nodeToNode.isStorage()) {
```
harbu: Wonder if we could remove this `if` so that tracker can assign any node to a stream? Up to tracker to determine logic?

src/logic/OverlayTopology.js
```diff
@@ -21,6 +21,10 @@ class OverlayTopology {
         return nodeId in this.nodes
     }
 
+    getRandomNode() {
+        return this.shuffleArray(Object.keys(this.nodes)).slice(0, 1)[0]
```
harbu: Btw there is also
```js
return this.pickRandomElement(Object.keys(this.nodes))
```

src/logic/Tracker.js
```diff
@@ -3,34 +3,47 @@ const createDebug = require('debug')
...
+const { nodeTypes } = require('../protocol/PeerBook')
 
 const NEIGHBORS_PER_NODE = 4
 
 module.exports = class Tracker extends EventEmitter {
     constructor(id, trackerServer) {
         super()
+
         this.overlayPerStream = {} // streamKey => overlayTopology
+        this.storages = new Map()
```
harbu: I think this can get confused with `Storage` interface in Node. What about name `storageNodes` to make it perfectly clear?

src/logic/Tracker.js
```diff
@@ -3,34 +3,47 @@ const createDebug = require('debug')
...
+        }
+
         this._updateNode(source, status.streams)
         this._formAndSendInstructions(source, status.streams)
+        this._formAndSendInstructionsToStorages()
     }
 
-    onNodeDisconnected(node) {
+    onNodeDisconnected(node, nodeType) {
+        if (nodeType === nodeTypes.STORAGE) {
```
harbu: I think this `if` is not strictly needed. `Map#delete(element)` will not fail if element does not exist In Map.

src/protocol/PeerBook.js
```diff
@@ -11,6 +11,12 @@ class MetadataNotSetError extends Error {
     }
 }
 
+const nodeTypes = Object.freeze({
```
harbu: Rename `peerTypes` for consistency?

src/protocol/PeerBook.js
```diff
@@ -60,21 +66,28 @@ class PeerBook {
...
     isTracker(peerId) {
-        if (!this.idToType[peerId]) {
-            throw new NotFoundInPeerBookError(`Id ${peerId} not found in peer book`)
-        }
-        return this.idToType[peerId] === 'tracker'
+        return this.getTypeById(peerId) === nodeTypes.TRACKER
     }
 
     isNode(peerId) {
+        return this.getTypeById(peerId) === nodeTypes.NODE
```
harbu: Should we also check for `STORAGE` here since storage nodes are also nodes?
I.e.
```js
return this.getTypeById(peerId) === nodeTypes.NODE || isStorage(peerId) 
```

src/protocol/TrackerServer.js
```diff
@@ -35,21 +35,31 @@ class TrackerServer extends EventEmitter {
...
+            })
         }
     }
 
     onMessageReceived(message) {
+        const nodeType = this.peerBook.getTypeById(message.getSource())
         switch (message.getCode()) {
             case encoder.STATUS:
-                this.emit(events.NODE_STATUS_RECEIVED, message)
+                this.emit(events.NODE_STATUS_RECEIVED, {
```
harbu, line 60: Because of lack of typing in JavaScript, I would prefer keeping `statusMessage` name here because it is more clear what the type of the message is. Especially in tests below.

src/logic/Tracker.js
```diff
@@ -85,6 +98,39 @@ module.exports = class Tracker extends EventEmitter {
...
     }
 
+    _formAndSendInstructionsToStorages() {
+        const existingStreams = Object.keys(this.overlayPerStream)
+
+        if (existingStreams.length) {
+            let streamsToSubscribe
+
+            this.storages.forEach(async (status, storageNode) => {
+                const alreadyConnected = Object.keys(status.streams)
```
harbu, line 109: What about re-naming to `alreadySubscribedStreams` etc. for consistency with variables `existingStreams` and `streamsToSubscribe`?

src/logic/Tracker.js
```diff
@@ -85,6 +98,39 @@ module.exports = class Tracker extends EventEmitter {
...
+                            try {
+                                await this.protocols.trackerServer.sendInstruction(storageNode, StreamID.fromKey(streamKey), [randomNode])
+                                this.debug('sent instruction %j for stream %s to storage node %s', [randomNode], streamKey, storageNode)
+                            } catch (e) {
+                                this.debug('failed to send instruction %j for stream %s to storage node %s because of %s', [randomNode], streamKey, storageNode, e)
+                            }
+                        }
+                    })
+                }
+            }, this)
```
harbu: No need for `this` with arrow function?

src/logic/Tracker.js
```diff
@@ -85,6 +98,39 @@ module.exports = class Tracker extends EventEmitter {
...
+            this.storages.forEach(async (status, storageNode) => {
+                const alreadyConnected = Object.keys(status.streams)
+
+                if (!alreadyConnected.length) {
+                    streamsToSubscribe = existingStreams
+                } else {
+                    streamsToSubscribe = existingStreams.filter((x) => !alreadyConnected.includes(x))
+                }
+
+                if (streamsToSubscribe.length) {
```
harbu, line 117: No need for `if` since `forEach` will not do anything if array is empty.

src/logic/Tracker.js
```diff
@@ -3,34 +3,47 @@ const createDebug = require('debug')
...
         const source = statusMessage.getSource()
         const status = statusMessage.getStatus()
+
+        if (nodeType === nodeTypes.STORAGE) {
+            this.storages.set(source, status)
+        }
+
         this._updateNode(source, status.streams)
         this._formAndSendInstructions(source, status.streams)
+        this._formAndSendInstructionsToStorages()
```
harbu, line 43: Btw is it necessary do it for all storage nodes at this point or just the storage node that sent status?

test/integration/storage-node.test.js
```diff
@@ -0,0 +1,82 @@
+const { startNetworkNode, startTracker, startStorageNode } = require('../../src/composition')
+const TrackerNode = require('../../src/protocol/TrackerNode')
+const { wait } = require('../util')
+const { callbackToPromise } = require('../../src/util')
+const { LOCALHOST, DEFAULT_TIMEOUT, waitForEvent } = require('../util')
+const { StreamID } = require('../../src/identifiers')
+
+jest.setTimeout(DEFAULT_TIMEOUT)
+
+describe('Check tracker will subscribe storage to all streams', () => {
```
harbu: `storage` => `storage node` in description

src/logic/Tracker.js
harbu: Answer to my own question: _it is necessary_ for any new streams that a non-storage node could register to Tracker via `_updateNode`.

test/integration/storage-node.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+        done()
+    })
+
+    it('tracker should register storage node and send subscribe all existing streams', async (done) => {
+        expect(tracker.storages.has('storage-1')).toEqual(false)
+
+        await subscriberOne.addBootstrapTracker(tracker.getAddress())
+        await subscriberTwo.addBootstrapTracker(tracker.getAddress())
+
```
harbu: Maybe add some `await` in between to make sure subscriberOne and subscriberTwo are registered before storageNode.

test/integration/storage-node.test.js
```diff
@@ -0,0 +1,82 @@
...
+        await callbackToPromise(tracker.stop.bind(tracker))
+        await callbackToPromise(subscriberOne.stop.bind(subscriberOne))
+        await callbackToPromise(subscriberTwo.stop.bind(subscriberTwo))
+        await wait(1000)
+    })
+
+    it('tracker should register storage node and send subscribe all new streams', async (done) => {
+        expect(tracker.storages.has('storage-1')).toEqual(false)
+
+        await storageNode.addBootstrapTracker(tracker.getAddress())
```
harbu: Btw `await` is not needed with `addBootstrapTracker` since it no longer returns promise or is async.

test/integration/storage-node.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+    it('tracker should register storage node and send subscribe all existing streams', async (done) => {
+        expect(tracker.storages.has('storage-1')).toEqual(false)
+
+        await subscriberOne.addBootstrapTracker(tracker.getAddress())
+        await subscriberTwo.addBootstrapTracker(tracker.getAddress())
+
+        await storageNode.addBootstrapTracker(tracker.getAddress())
+
+        let instuctionsReceived = 0
```
harbu: instuctions => instructions, missing r between

test/integration/storage-node.test.js
```diff
@@ -0,0 +1,82 @@
...
+        storageNode.protocols.trackerNode.on(TrackerNode.events.TRACKER_INSTRUCTION_RECEIVED, async (streamMessage) => {
+            instuctionsReceived += 1
+            const streamId = streamMessage.getStreamId()
+            const nodeAddresses = streamMessage.getNodeAddresses()
+
+            if (streamId.key() === 'stream-1::0') {
```
harbu: I thinks these checks are not strictly necessary because the assertions of `getAllNodesForStream` below are also derived from addresses in protocol layer.

harbu (changes requested):
Overall looks good. I have some nitpicks and then a more bigger question about the `_formAndSendInstructionsToStorages` method.

test/integration/storage-node.test.js
mirpo: Good point 👍

src/logic/Node.js
mirpo: Hmmm, I tested. It works without this if, now I'm trying to image any situation which will break code. Removed for now.

src/protocol/PeerBook.js
harbu: @mirpo what about this one?

harbu (approved):

PR #233, opened 2019-04-09
L1 Resend Support

harbu:
Network now supports resend at L1, which means serving clients directly from a network node's local storage. Proxying resend requests to other nodes (L2) and to storage nodes (L3) are the next steps and come later.

## Summary of changes

- Class `ResendHandler` handles resend requests and interactions between `Node` and `Storage`.
- Object implementing `Storage` "interface" is now passed to `Node` via constructor.
- Object implementing `Storage` "interface" can be passed to `startNode` and `startNetworkNode` in composition.js. 
- `NoOpStorage` is the default (and currently only) class implementing `Storage`. As its name implies, it doesn't store anything and always returns empty streams.
- Add convenience methods to `NetworkNode` for requesting resends. Will be used by [broker](https://github.com/streamr-dev/broker) on behalf of clients.
- Made integration test `l1-resend-requests.test.js` to check that resend requests works at L1 level, end-to-end.

### DIFF
--- a/src/NetworkNode.js
+++ b/src/NetworkNode.js
@@ -1,4 +1,7 @@
 const DataMessage = require('./messages/DataMessage')
+const ResendLastRequest = require('./messages/ResendLastRequest')
+const ResendFromRequest = require('./messages/ResendFromRequest')
+const ResendRangeRequest = require('./messages/ResendRangeRequest')
 const Node = require('./logic/Node')
 const { StreamID, MessageID, MessageReference } = require('./identifiers')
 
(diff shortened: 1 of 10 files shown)


src/composition.js
```diff
@@ -20,25 +21,25 @@ async function startTracker(host, port, id = uuidv4()) {
     })
 }
 
-async function startNode(host, port, id = uuidv4()) {
+async function startNode(host, port, id = uuidv4(), storage = new NoOpStorage()) {
```
mirpo, line 24: storage should be an array, for fallback MemDB -> Cassandra -> etc, but could be done later in separate ticket

src/NetworkNode.js
```diff
@@ -54,4 +57,36 @@ module.exports = class NetworkNode extends Node {
     unsubscribe(streamId, streamPartition) {
         this.unsubscribeFromStream(new StreamID(streamId, streamPartition))
     }
+
```
mirpo, line 60: We else need support of subscribe with resend

src/logic/Node.js
```diff
@@ -31,6 +37,8 @@ class Node extends EventEmitter {
             this.debug('failed to deliver buffered messages of stream %s', streamId)
             this.emit(events.MESSAGE_DELIVERY_FAILED, streamId)
         })
+        this.resendHandler = new ResendHandler(storage)
+        this._bindResendHandlerEventsToNodeEvents()
```
mirpo, line 41: What if you are running publisher node/realtime node? Maybe we should bind events only even if we initiated ResendHandler/storage and report that "sorry, no storage support" otherwise and propagate resend request father.

test/unit/ResendHandler.test.js
```diff
@@ -0,0 +1,202 @@
...
+    describe('after receiving resend request', () => {
+        let events
+        let doRequest
+
+        beforeEach(() => {
+            events = eventsToArray(resendHandler, Object.values(ResendHandler.events))
+            doRequest = async () => {
+                resendHandler.handleRequest(new ResendLastRequest(new StreamID('streamId', 0), 'subId', 10, 'source'))
+
+                // Wait for 5 events
```
mirpo, line 72: events?

mirpo (approved):
LGFM

PR #245, opened 2019-04-11
Refactor resend handler to use strategy pattern

### LINKED ISSUE #235: Multiple storages on Node
Ability to pass multiple storages to Node instead of one. E.g.
```js
startNetworkNode('127.0.0.1', 3800, ..., [inMemoryStorage, cassandraStorage])
```

harbu:
While starting to work on L2 resend requests, I noticed that the `ResendHandler` class was becoming large and confusing. I noticed then that we can actually refactor the `ResendHandler` to take an array of "resend strategies". It will try them one-by-one until one of them fulfills the resend request and then it short-circuits. The added benefit is that we can configure different strategies, order of strategies, or even re-use strategies with different parameters.

So, usually we would have something like, e.g.,
```js
startNode(host, port, id, [L1strategy, L2strategy, L3strategy])
```

or for multiple storage nodes (fix #235)

```js
startNode(host, port, id, [new StorageResendStrategy(redisStorage), new StorageResendStrategy(cassandraStorage)])
```

### Summary of changes
- `startNetworkNode(...)` as last argument takes list of storages instead of one storage
- `startNode(..)` takes as last argument list of resend strategies instead of `ResendHandler`.
- Remove `NoOpStorage` since same functionality is provided by `ResendHandler` given no strategies.
- `ResendHandler` takes care of executing resend strategies one-by-one until one of them fulfills the request (RESENT) or none of them do (NO_RESEND).
- Replaced events of `ResendHandler` with callback functions given as arguments in constructor. 
- Storage resend logic was moved from `ResendHandler` to `StorageResendStrategy` residing in resendStrategies.js


src/logic/Node.js
```diff
@@ -37,8 +35,17 @@ class Node extends EventEmitter {
             this.debug('failed to deliver buffered messages of stream %s', streamId)
             this.emit(events.MESSAGE_DELIVERY_FAILED, streamId)
         })
-        this.resendHandler = resendHandler
-        this._bindResendHandlerEventsToNodeEvents()
+        this.resendHandler = new ResendHandler(resendStrategies, this.respondResend.bind(this),
```
mirpo, line 38: Why not to create and configure ResendHandler out of constructor in composition.js?

mirpo (approved):
LGFM

src/logic/Node.js
harbu: Could also instantiated in composition.js. But configuring would be harder since we need to pass it callback functions as arguments and some of them are internal to class `Node`.

PR #247, opened 2019-04-13
Fix subscribe flow

mirpo:


src/logic/Node.js
```diff
@@ -205,18 +203,23 @@ class Node extends EventEmitter {
             source
         })
 
-        this.subscribeToStreamIfHaveNotYet(streamId)
+        if (this.streams.isSetUp(streamId)) {
```
harbu, line 206: Nice

src/logic/Node.js
```diff
@@ -205,18 +203,23 @@ class Node extends EventEmitter {
...
-            source
-        })
+            this.debug('node %s subscribed to stream %s', source, streamId)
+            this.emit(events.NODE_SUBSCRIBED, {
+                streamId,
+                source
+            })
+        } else {
+            this.debug('node %s tried to subscribe to stream %s, but it is not setup', source, streamId)
+            this.protocols.nodeToNode.disconnectFromNode(source, disconnectionReasons.NO_SHARED_STREAMS)
```
harbu, line 221: At this point we could have other shared streams with `source`:

1) A new disconnectionReason for this could be more clearer?

2) Disconnecting will remove any existing shared streams the the two nodes may have. This will cause the network to need to be re-structured, which is a bit wasteful. But I understand that this is the only solution now because there is no other mechanism to deny the incoming subscription. I think we need a new issue for a new message type, e.g., `SUBSCRIBE_RESPONSE` that can be used to confirm or deny subscription request? What do you think?

harbu (approved):
LGTM; I like the solution. See comment for new issue idea

src/logic/Node.js
mirpo: I agree, we need -> SUBSCRIBE_RESPONSE 👍 I tested running network 500+ it works stable, but ofc, we don't test now multiple streams in network, that could cause some issues.

PR #255, opened 2019-04-15
travis.yml: try all scripts in one stage (WIP)

harbu:


mirpo:
Conflicting

harbu:
Never mind. This does make tests run considerably faster. But it also makes interpreting test failures harder when looking at Travis build status page. Current turnaround time of Travis is good enough (~2 min) so there is probably more downsides that benefits to this change.

PR #258, opened 2019-04-18
In MemoryStorage

mirpo:
We should decide how we want to use streams in pull or push mode.


src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,105 @@
+const { Readable } = require('stream')
+
+module.exports = class MemoryStorage {
+    constructor() {
+        this.storage = new Map()
+        this.index = new Map()
+    }
+
+    store(streamId, streamPartition, subId, timestamp, sequenceNo, publisherId, msgChainId, data) {
```
harbu: Missing a few fields and `subId` not necessary. Also `store()` is passed an object so should use `store({ ... })`. Check this for example:
https://github.com/streamr-dev/network/blob/96d9018a16c89cc92b182599c0ba7c76e8d167e1/src/NetworkNode.js#L104-L116

src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,105 @@
...
+        return this.hasStreamId(streamId, streamPartition, subId) ? Object.keys(this.storage.get(index)).length : 0
+    }
+
+    _createStream(fetchFunc, index, streamId, streamPartition, subId) {
+        const stream = new Readable({
+            objectMode: true
+        })
+
+        // eslint-disable-next-line no-underscore-dangle
+        stream._read = () => {
```
harbu: I believe if you pass this function in the constructor of Readable you can avoid the underscore-dangle, e.g.,
```js
const stream = new Readable({
    objectMode: true,
    read() { 
       ...
    }
  })
```

src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,105 @@
...
+            throw new TypeError('number is not an positive integer')
+        }
+
+        const index = this._index(streamId, streamPartition, subId)
+        const filterFunc = () => this.index.get(index).slice(-number)
+
+        return this._createStream(filterFunc, index, streamId, streamPartition, subId)
+    }
+
+    from(streamId, streamPartition, subId, fromTimestamp, fromSequenceNo = '', publisherId = '') {
```
harbu: Parameter `fromSequenceNo` should be an integer? Also maybe worth validating below as done for `fromTimestamp`.

src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,105 @@
...
+        if (!Number.isInteger(fromTimestamp) || fromTimestamp <= 0) {
+            throw new TypeError('fromTimestamp is not an positive integer')
+        }
+
+        const index = this._index(streamId, streamPartition, subId)
+        const filterFunc = () => this.index.get(index).filter((id, timestamp) => timestamp >= fromTimestamp)
```
harbu: Should also filter by `fromSequenceNo` and `publisherId` if given to function.

src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,105 @@
...
+
+        return this._createStream(filterFunc, index, streamId, streamPartition, subId)
+    }
+
+    range(streamId, streamPartition, subId, fromTimestamp, toTimestamp, fromSequenceNo = 0, toSequenceNo = 0, publisherId = '') {
```
harbu: Parameters `fromSequence` and `toSequence` should be integers.

src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,105 @@
...
+        if (!Number.isInteger(toTimestamp) || toTimestamp <= 0) {
+            throw new TypeError('toTimestamp is not an positive integer')
+        }
+
+        if (fromTimestamp > toTimestamp) {
+            throw new TypeError('fromTimestamp must be less or equal than toTimestamp')
+        }
+
+        const index = this._index(streamId, streamPartition, subId)
+        const filterFunc = () => this.index.get(index).filter((id, timestamp) => timestamp >= fromTimestamp && timestamp <= toTimestamp)
```
harbu: Should also filter by `fromSequenceNo`, `toSequenceNo`, and `publisherId` if given.

test/unit/MemoryStorage.test.js
```diff
@@ -0,0 +1,149 @@
...
+                }
+            }
+        })
+
+        last10Stream.on('end', () => {
+            expect(arr.length).toEqual(RECORDS_NEEDED)
+
+            const data = dataMessages.slice(-RECORDS_NEEDED).map((dataMessage) => dataMessage.getData())
+
+            expect(arr.length).toEqual(RECORDS_NEEDED)
```
harbu: Same assertion twice

test/unit/MemoryStorage.test.js
```diff
@@ -0,0 +1,149 @@
+const MemoryStorage = require('../../src/storage/MemoryStorage')
+const DataMessage = require('../../src/messages/DataMessage')
+const { StreamID, MessageID, MessageReference } = require('../../src/identifiers')
+
+const dataMessages = []
+const dataMessages2 = []
+const MAX = 100000
```
harbu: Why so many messages? Wouldn't less be enough for unit test?

src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,105 @@
...
+
+module.exports = class MemoryStorage {
+    constructor() {
+        this.storage = new Map()
+        this.index = new Map()
+    }
+
+    store(streamId, streamPartition, subId, timestamp, sequenceNo, publisherId, msgChainId, data) {
+        const id = `${timestamp}, ${sequenceNo}, ${publisherId}, ${msgChainId}`
+        const index = this._index(streamId, streamPartition, subId)
```
harbu: I continually mixed up `id`, `index`, and `this.index`  in my head when reading. More descriptive names would be helpful.

test/unit/MemoryStorage.test.js
```diff
@@ -0,0 +1,149 @@
...
+                if (data !== null) {
+                    arr.push(data)
+                } else {
+                    break
+                }
+            }
+        })
+
+        fromStream.on('end', () => {
+            const dataMessagesNeeded = dataMessages.filter((dataMessage) => dataMessage.getMessageId().timestamp >= FROM_TIME && dataMessage.getMessageId().timestamp <= TO_TIME)
```
harbu: Each of these test cases repeat the filtering condition of the method under test. I personally prefer a style where we check more concrete values. For example:

1. Create only 10 messages with timestamp 0 to 9.
2. Make, e.g., `FROM_TIME = 5` and `TO_TIME=7`
3. `expect(arr.length).toEqual(3)` (You can even skip this one since this is implied by 4.  )
4. `expect(arr.data).toEqual([{ messageNo: 5 }, { messageNo: 6 }, { messageNo: 7 }])`

harbu (changes requested):
Some general points

- No need for `subId` in this class. `subId` is used for resend logic, to keep track of request "session". But storage need not know anything about it.

- Currently using push stream model in `ResendHandler` and resend strategies. I guess this code is using pull stream model from the looks of it?

- Chunks of ReadableStream should contain the following fields of `DataMessage` (not only `data`):              timestamp, sequenceNo,  publisherId, msgChainId, previousTimestamp, previousSequenceNo, data, signature, signatureType, . See here: https://github.com/streamr-dev/network/blob/96d9018a16c89cc92b182599c0ba7c76e8d167e1/src/logic/resendStrategies.js#L8-L33

- Naming of functions `last` => `requestLast`, `from` => `requestFrom`, and `range` => `requestRange`. See here for more: https://github.com/streamr-dev/network/blob/master/src/logic/resendStrategies.js#L47-L80

- We probably need some size limit & eviction strategy  for `this.storage` and `this.index`, otherwise we will run out of memory when enough messages have been received.

src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,124 @@
...
+            throw new TypeError('fromTimestamp is not an positive integer')
+        }
+
+        if (!Number.isInteger(fromSequenceNo) || fromSequenceNo < 0) {
+            throw new TypeError('fromSequenceNo should be equal or greater than zero')
+        }
+
+        const streamKey = this._getStreamKey(streamId, streamPartition)
+        const filterFunc = () => Object.values(this.storage.get(streamKey)).filter((record) => {
+            return record.timestamp >= fromTimestamp && record.sequenceNo === fromSequenceNo
```
harbu: Two points:

1. When `publisherId === null` it shouldn't be used in filtering so its condition should be
```js
publisherId === null || record.publisherId === publisherId
```

2. Pair `(fromTimestamp, fromSequenceNo)` denote starting point of "from query". So the condition should be something like

```js
record.timestamp > fromTimestamp || (record.timestamp === fromTimestamp && record.sequenceNo >= fromSequenceNo)
```

src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,124 @@
...
+
+        if (!Number.isInteger(fromSequenceNo) || fromSequenceNo < 0) {
+            throw new TypeError('fromSequenceNo should be equal or greater than zero')
+        }
+
+        if (!Number.isInteger(toSequenceNo) || toSequenceNo < 0) {
+            throw new TypeError('toSequenceNo should be equal or greater than zero')
+        }
+
+        if (fromSequenceNo > toSequenceNo) {
```
harbu: Not necessarily. It is fine to have `fromSequenceNo > toSequenceNo` when `fromTimestamp != toTimestamp`. E.g. `(1556017275, 500)` and `(1556019999, 0)`.

src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,124 @@
...
+            throw new TypeError('fromSequenceNo must be less or equal than toSequenceNo')
+        }
+
+        const streamKey = this._getStreamKey(streamId, streamPartition)
+        const filterFunc = () => Object.values(this.storage.get(streamKey)).filter((record) => {
+            return record.timestamp >= fromTimestamp && record.timestamp <= toTimestamp
```
harbu: Conditions need to be changed to the same effect as the comment on `requestFrom`. Here we additionally need to check upper bound, so messages from `(fromTimestamp, fromSequenceNo)` up to `(toTimestamp, toSequenceNo)`.

src/storage/MemoryStorage.js
```diff
@@ -0,0 +1,124 @@
+const { Readable } = require('stream')
+
+module.exports = class MemoryStorage {
+    constructor(maxNumberOfMessages = 10000) {
+        this.maxNumberOfMessages = maxNumberOfMessages
+        this.storage = new Map()
+    }
+
+    store({
+        streamId, streamPartition, timestamp, sequenceNo, publisherId, msgChainId, previousSequenceNo, data, signature, signatureType
```
harbu: Still need `previousTimestamp`

harbu (changes requested):

harbu (approved):

PR #259, opened 2019-04-22
L3 resend requests

harbu:
- Extract proxying resend logic from  `AskNeighborsResendStrategy` to own class `ProxiedResend` because same behavior is needed in L3.
- Add two new message types that are required for L3:
    - `FindStorageNodesMessage` is sent from node to tracker
    - tracker responds with `StorageNodesMessage` containing storage nodes.
- Method `Tracker#findStorageNodes` that handles responding to `FindStorageNodesMessage` with `StorageNodesMessage`.
- The actual L3 implementation in `StorageNodeResendStrategy`. Quite heavy logic, might want to go thru face-to-face. Also large unit test for `StorageNodeResendStrategy`.
- Add L3 to `NetworkNode`
- Integration test `l3-resend-requests.test.js` for testing l3 with real configuration.


harbu:
Close and make new PR so Travis will run it.

PR #267, opened 2019-04-25
Global bin

mirpo:
- Added support for global installation
- Added basic pubsub example
- Rewritten readme, preparing for the release

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,102 +1,108 @@
-# Streamr P2P network  ![Travis](https://travis-ci.com/streamr-dev/network.svg?token=qNNVCnYJo1fz18VTNpPZ&branch=master)
-
-# Installation
-
-### Mac OS
-
-#### install brew 
-`/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
-
-#### install nvm
-
-`brew update`
-
-`brew doctor`
+# Streamr network 
+ ![Travis](https://travis-ci.com/streamr-dev/network.svg?token=qNNVCnYJo1fz18VTNpPZ&branch=master)
  
-`brew install nvm`
+> P2P network for the real-time data with storage support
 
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -54,5 +61,6 @@
     "into-stream": "^5.1.0",
     "jest": "^24.7.1",
     "p-event": "^4.1.0"
-  }
+  },
+  "preferGlobal": true
```
harbu: This is deprecated according to https://docs.npmjs.com/files/package.json#preferglobal

Also there are two ways to use this npm package
- To install for bin commands
- To install as library dependency (e.g. broker)

So we shouldn't necessarily warn if someone just wants to use this as library.

examples/pubsub/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "mqqt",
```
harbu: update name

README.md
```diff
@@ -1,102 +1,108 @@
...
-`npm run test-integration`
+    npm run coverage
+    
+Debugging:
 
-code coverage
+To get all Streamr Network debug messages  
 
-`./node_modules/jest/bin/jest.js --coverage`
+    export DEBUG=streamr:*
```
harbu, line 98: 👍

README.md
```diff
@@ -1,102 +1,108 @@
...
 
-run one test
+- connection layer `export DEBUG=streamr:connection:*`
+- logic layer `export DEBUG=streamr:logic:*`
+- protocol layer `export DEBUG=streamr:protocol:*`
 
-`./node_modules/jest/bin/jest.js ./test/integration/message-duplication.test.js`
+Excluding level
 
-# Releasing
```
harbu, line 96: Was this removed on purpose? Useful for us at least for releasing because there are some steps involved.

harbu (changes requested):

harbu (approved):

PR #277, opened 2019-05-24
Updated ws 7.0

mirpo:
* Updated ws to 7.0
  * We can't remove or additional check for duplicates and etc
* Updated duplicate-connections-are-closed.test.js
  * It's still flaky if you run test in loop (100 times), but I think it caused that sockets are not closed fully on OS level.


harbu (approved):
LGTM!

Issue #261, opened 2019-04-24
Parameters


harbu:
Make timeouts, intervals, and other limits configurable.

PR #278, opened 2019-05-27
Parameters in constructors

mirpo:
I moved all possible parameters/intervals to some options object inside classes (#261)
I decided to keep `this.protocols = this.opts.protocols` to be short in out listeners `this.protocols.trackerNode.on`

### DIFF
--- a/src/logic/Node.js
+++ b/src/logic/Node.js
@@ -23,15 +23,24 @@ const events = Object.freeze({
 const MIN_NUM_OF_OUTBOUND_NODES_FOR_PROPAGATION = 1
 
 class Node extends EventEmitter {
-    constructor(id, trackerNode, nodeToNode, resendStrategies) {
+    constructor(trackerNode, nodeToNode, resendStrategies, opts) {
         super()
 
-        this.connectToBoostrapTrackersInterval = setInterval(this._connectToBootstrapTrackers.bind(this), 5000)
+        // set default options
+        const defaultOptions = {
+            id: 'node',
+            connectToBootstrapTrackersInterval: 5000,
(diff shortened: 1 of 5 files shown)


src/logic/Node.js
```diff
@@ -23,30 +23,37 @@ const events = Object.freeze({
 const MIN_NUM_OF_OUTBOUND_NODES_FOR_PROPAGATION = 1
 
 class Node extends EventEmitter {
-    constructor(id, trackerNode, nodeToNode, resendStrategies) {
+    constructor(opts) {
         super()
 
-        this.connectToBoostrapTrackersInterval = setInterval(this._connectToBootstrapTrackers.bind(this), 5000)
+        // set default options
+        const defaultOptions = {
```
harbu, line 30: Could be worth validating that `protocol.trackerNode` and `protocol.nodeToNode` are given since this class won't work without them set?

src/logic/Tracker.js
```diff
@@ -6,22 +6,26 @@ const { StreamID } = require('../identifiers')
...
+        if (!Number.isInteger(opts.maxNeighborsPerNode)) {
+            throw new Error('maxNeighborsPerNode is not an integer')
         }
 
-        if (!Number.isInteger(maxNeighborsPerNode)) {
-            throw new Error('maxNeighborsPerNode is not an integer')
+        // set default options
+        const defaultOptions = {
+            id: 'tracker',
+            protocols: []
```
harbu, line 19: Same point as before.

harbu (approved):
LGTM; couple minor validations that could be added

PR #279, opened 2019-06-07
Use streamr client protocol js

mthambipillai:
This PR replaces the existing protocol implementation for the Control Layer and the Message Layer with the protocol implementation used in the centralised network and defined [here](https://github.com/streamr-dev/streamr-client-protocol-js).


.idea/runConfigurations/integration_test.xml
```diff
@@ -1,12 +1,13 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="integration test" type="JavaScriptTestRunnerJest">
+  <configuration default="false" name="integration test" type="JavaScriptTestRunnerJest" factoryName="Jest">
     <config-file value="$PROJECT_DIR$/jest.config.js" />
-    <node-interpreter value="project" />
+    <node-interpreter value="$USER_HOME$/.nvm/versions/node/v10.12.0/bin/node" />
```
mirpo: Not sure, is it needed?

.idea/runConfigurations/test.xml
```diff
@@ -1,9 +1,8 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="test" type="JavaScriptTestRunnerJest">
+  <configuration default="false" name="test" type="JavaScriptTestRunnerJest" factoryName="Jest">
     <config-file value="$PROJECT_DIR$/jest.config.js" />
-    <node-interpreter value="project" />
+    <node-interpreter value="$USER_HOME$/.nvm/versions/node/v10.12.0/bin/node" />
```
mirpo: The same

src/logic/Node.js
```diff
@@ -93,9 +94,9 @@ class Node extends EventEmitter {
     }
 
     subscribeToStreamIfHaveNotYet(streamId) {
-        if (!this.streams.isSetUp(streamId)) {
+        if (!this.streams.isSetUp(streamId, streamId)) {
```
mirpo: isSetUp - takes only one parameter

src/logic/Node.js
```diff
@@ -93,9 +94,9 @@ class Node extends EventEmitter {
...
             this.debug('add %s to streams', streamId)
-            this.streams.setUpStream(streamId)
+            this.streams.setUpStream(streamId, streamId)
```
mirpo: setUpStream - takes only one parameter

src/logic/Node.js
```diff
@@ -173,72 +174,61 @@ class Node extends EventEmitter {
         })
     }
 
-    onDataReceived(dataMessage) {
-        const messageId = dataMessage.getMessageId()
-        const previousMessageReference = dataMessage.getPreviousMessageReference()
-        const { streamId } = messageId
+    onDataReceived(streamMessage, source = null) {
```
mirpo, line 177: Is it possible that source could be null? @harbu

mirpo (changes requested):
LGFM 👍

mirpo:
One point about bringing JS protocol into network, I don't like idea of removing "source" from messages. Each message in network supposed to have a source, so it is atomic structure - "message + source".
So I think it's better to add Network implementation of each message in protocol (aka extend and add fields) or extend each message class in network project.

.idea/misc.xml
```diff
@@ -3,5 +3,5 @@
   <component name="JavaScriptSettings">
     <option name="languageLevel" value="ES6" />
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_6" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_6" default="false" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
```
harbu: Are these project file changes purposeful?

src/NetworkNode.js
```diff
@@ -87,14 +83,15 @@ class NetworkNode extends Node {
         content,
         signature,
```
harbu: Might be worth changing the API of publish so that `signatureType` is before `signature` for consistency.

src/NetworkNode.js
```diff
@@ -21,30 +17,30 @@ const events = Object.freeze({
 })
 
 function unicastMessageToObject(unicastMessage) {
```
harbu: Let's discuss in next meeting whether we want to evolve this API to directly return streamr-client-protocol-js messages, or is the current just plain objects solution fine.

src/helpers/MessageEncoder.js
```diff
@@ -29,6 +20,9 @@ const encode = (type, payload) => {
 
 const decode = (source, message) => {
     const { code, payload } = JSON.parse(message)
+    if (code === undefined) {
```
harbu: Do we need a _protocol layer_ in our network anymore? Or is `streamr-client-protocol-js` basically the protocol layer. Let's discuss in next meeting.

src/logic/Node.js
harbu: It is `null` when the data is being published locally.

src/logic/Node.js
```diff
@@ -173,72 +174,61 @@ class Node extends EventEmitter {
...
+                this._propagateMessage(streamMessage, source)
             } else {
-                this.debug('ignoring duplicate data %s (from %s)', messageId, dataMessage.getSource())
+                this.debug('ignoring duplicate data %s (from %s)', streamMessage.messageId, source)
                 this.metrics.received.duplicates += 1
             }
         } else {
             this.debug('Not outbound nodes to propagate')
-            this.messageBuffer.put(streamId.key(), dataMessage)
+            this.messageBuffer.put(`${streamMessage.getStreamId()}::${streamMessage.getStreamPartition()}`, [streamMessage, source])
```
harbu: You can use `streamIdAndPartition.key()` to avoid repeating the key forming logic.

src/logic/Node.js
```diff
@@ -173,72 +174,61 @@ class Node extends EventEmitter {
...
+            this.seenButNotPropagated.delete(streamMessage.messageId)
+            this.emit(events.MESSAGE_PROPAGATED, streamMessage)
         } else {
             // Handle scenario in which we were unable to propagate message to enough nodes. This often happens when
             // socket.readyState=2 (closing)
             this.debug('put %s back to buffer because could not propagated %d nodes or more',
-                messageId, MIN_NUM_OF_OUTBOUND_NODES_FOR_PROPAGATION)
-            this.seenButNotPropagated.add(messageId)
-            this.messageBuffer.put(streamId.key(), dataMessage)
+                streamMessage.m, MIN_NUM_OF_OUTBOUND_NODES_FOR_PROPAGATION)
```
harbu: `streamMessage.messageId` ?

src/logic/Node.js
```diff
@@ -173,72 +174,61 @@ class Node extends EventEmitter {
...
+            this.seenButNotPropagated.add(streamMessage.messageId)
+            this.messageBuffer.put(`${streamMessage.getStreamId()}::${streamMessage.getStreamPartition()}`, [streamMessage, source])
```
harbu: Same point as before regarding `key()` method

src/protocol/NodeToNode.js
```diff
@@ -37,66 +40,16 @@ class NodeToNode extends EventEmitter {
...
 
-    respondResent(receiverNodeId, streamId, subId) {
-        const receiverNodeAddress = this.peerBook.getAddress(receiverNodeId)
-        return this.endpoint.send(receiverNodeAddress, encoder.resendResponseResent(streamId, subId))
-    }
-
-    respondNoResend(receiverNodeId, streamId, subId) {
-        const receiverNodeAddress = this.peerBook.getAddress(receiverNodeId)
-        return this.endpoint.send(receiverNodeAddress, encoder.resendResponseNoResend(streamId, subId))
+    sendUnsubscribe(receiverNodeId, streamIdAndPartition) {
```
harbu, line 53: Just a thought: would it be worth renaming the `StreamId` class to `StreamIdAndPartition`? It would be more line in with the naming convention going on in this PR's change set.

src/protocol/NodeToNode.js
```diff
@@ -106,66 +59,9 @@ class NodeToNode extends EventEmitter {
...
-                message.getMessageId(),
-                message.getPreviousMessageReference(),
-                message.getData(),
-                message.getSignature(),
-                message.getSignatureType(),
-                message.getSubId()
-            )
-        }
-        throw new Error(`unrecognized message ${message}`)
+    send(receiverNodeId, message) {
```
harbu, line 64: 👍

src/protocol/TrackerNode.js
```diff
@@ -37,6 +36,10 @@ class TrackerNode extends EventEmitter {
     }
 
     onMessageReceived(message) {
+        // Unifying all the messages in the same protocol with a third layer would allow to avoid this hack
```
harbu: Not sure I understand this part. Why do we return on `message.version === 1`? Is it to prevent node-to-node messages from getting handled?

test/integration/storage-node.test.js
```diff
@@ -97,7 +97,7 @@ describe('Check tracker will subscribe storage node to all streams', () => {
             await waitForEvent(tracker.protocols.trackerServer, TrackerServer.events.NODE_STATUS_RECEIVED)
         ])
 
-        expect(subscriberOne.streams.isSetUp(streamOne)).toBeFalsy()
+        expect(subscriberOne.streams.isSetUp(streamOne, streamOne)).toBeFalsy()
```
harbu: two arguments?

harbu (changes requested):
LGTM; I added some comments of potential things I found + a few questsions + a few discussion topics for network meeting.

harbu:
> One point about bringing JS protocol into network, I don't like idea of removing "source" from messages. Each message in network supposed to have a source, so it is atomic structure - "message + source".
> So I think it's better to add Network implementation of each message in protocol (aka extend and add fields) or extend each message class in network project.

We could have some sort of a wrapper, e.g.,
```js
class WrappedMessage {
   constructor(msg, source) {
      this.msg = msg
      this.source = source
   }
}
```

Though a more lightweight (and perhaps a bit less 'Javaish' solution) would be to just pass objects
```js
{
   msg: msg,
   source: 'sourceId'
}
```

In this PR we pass arrays that have a positional interpretation
```js
[msg, source]
```

I think a shortcoming here is that we don't have a type system in use. That would catch situations where I'm trying to, e.g.,

```js 
const [sourceId, msg] = xxx
doSomething(msg.streamId)
```

even though it should be
```js
const [msg, source] = xxx
doSomething(msg.streamId)
```

src/helpers/MessageEncoder.js
mthambipillai: Yes it's still needed for messages that are not in `streamr-client-protocol-js`. This is why I was suggesting to add a third layer in the last meeting.

src/protocol/TrackerNode.js
mthambipillai: Yes this is the reason. This is why in the last meeting I was suggesting a third layer with a new message like "NodeToNodeMessage" that would encapsulate Control Layer payloads and some node specific info (like the source). It seems like your comment below suggests the same thing

harbu:
I checked new commits. Looks good to me. Only thing I would still like to see is tests for new wrapper message type in
- test/unit/encoder.test.js
- test/integration/delivery-of-messages-protocol-layer.test.js (if applicable)

mthambipillai:
Added unit tests in `encoder.test.js`. Didn't add tests in `delivery-of-messages-protocol-layer.test.js` since there already are tests for every possible payload of the WrapperMessage.

PR #283, opened 2019-06-20
add metrics: messageBuffer size & open handles

harbu:
Added two metrics that are useful as signals for node not being able to keep up with amount of publishing.


mirpo (approved):
👍

PR #285, opened 2019-07-02
simplify resend response stream

harbu:
Remove `unicastMessageSource` from resend response stream since the user of this library (i.e. broker) has no need for this information.

Simplifies items of response stream to be `UnicastMessage` instead of `[UnicastMessage, source]`.


mirpo (approved):
lgfm

PR #287, opened 2019-07-03
Add process cpu and memory usage metrics

mirpo:


src/metrics.js
```diff
@@ -1,16 +1,23 @@
...
         this._metrics = new Map()
+        this.pidusage = pidusage
     }
 
     createSpeedometer(name) {
         this._metrics.set(name, speedometer())
     }
 
+    async getPidusage(pid = process.pid) {
+        const stats = await this.pidusage(pid)
```
harbu: Why set `this.pidusage = pidusage` and not just use directly `pidusage(pid)` ?

src/logic/Node.js
```diff
@@ -377,9 +377,10 @@ class Node extends EventEmitter {
         }
     }
 
-    getMetrics() {
+    async getMetrics() {
         const metrics = this.protocols.nodeToNode.endpoint.getMetrics()
 
+        metrics.pidMetrics = await this.metrics.getPidusage()
```
harbu: Maybe move this line inside`Metrics#report()`. That way also tracker will report pid usage.

harbu (changes requested):

src/metrics.js
mirpo: 👍

src/logic/Node.js
mirpo: 👍

harbu (approved):

PR #288, opened 2019-07-12
Add topology to tracker metrics, update ws

mirpo:


src/logic/Tracker.js
```diff
@@ -174,6 +174,13 @@ module.exports = class Tracker extends EventEmitter {
         const trackerMetrics = this.metrics.report()
         const mainMetrics = this.metrics.prettify(endpointMetrics)
 
+        const topology = []
```
harbu:
```js
const topology = {}
```

harbu (approved):

PR #318, opened 2019-08-01
NetworkNode#publish take StreamMessage

### LINKED ISSUE #309: NetworkNode#publish should take StreamMessage as parameter
Currently method `NetworkNode#publish` takes a list of arguments that are used within the method to build a StreamMessage to pass along.
https://github.com/streamr-dev/network/blob/dbf15601a4c119dec0e91b152096cdba2de7c77e/src/NetworkNode.js#L46-L67

It should just take a `StreamMessage` directly so it would be consistent with the other public methods.

harbu:
Fixes #309 

Method `NetworkNode#publish` now takes a `StreamMessage` in its parameter. 

*Not* backwards compatible. Broker will need to be changed to accommodate this change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3383,6 +3383,11 @@
         "minimalistic-assert": "^1.0.0"
       }
     },
+    "heap": {
+      "version": "0.2.6",
+      "resolved": "https://registry.npmjs.org/heap/-/heap-0.2.6.tgz",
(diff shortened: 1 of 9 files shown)


harbu:
When https://github.com/streamr-dev/streamr-client-protocol-js/pull/15 is landed and published to NPM can shave a few lines off tests.

mirpo (approved):

PR #319, opened 2019-08-01
Tracker overlayPerStream test, remove correctly empty node from tracker

mirpo:
- Tracker overlayPerStream test, remove correctly empty node from tracker
- Remove empty overlay from tracker

### DIFF
--- a/src/logic/OverlayTopology.js
+++ b/src/logic/OverlayTopology.js
@@ -38,9 +38,12 @@ class OverlayTopology {
         }
     }
 
+    isEmpty() {
+        return Object.entries(this.nodes).length === 0
+    }
+
     state() {
-        const entries = Object.entries(this.nodes)
-        return entries.length ? Object.assign(...entries.map(([nodeId, neighbors]) => {
+        return !this.isEmpty() ? Object.assign(...Object.entries(this.nodes).map(([nodeId, neighbors]) => {
             return {
                 [nodeId]: [...neighbors].sort()
             }
--- a/src/logic/OverlayTopology.js
+++ b/src/logic/OverlayTopology.js
@@ -34,8 +34,8 @@ class OverlayTopology {
     leave(nodeId) {
         if (this.nodes[nodeId] != null) {
             this.nodes[nodeId].forEach((neighbor) => this.nodes[neighbor].delete(nodeId))
+            delete this.nodes[nodeId]
         }
-        delete this.nodes[nodeId]
     }
 
     isEmpty() {
--- a/src/logic/OverlayTopology.js
+++ b/src/logic/OverlayTopology.js
@@ -34,8 +34,8 @@ class OverlayTopology {
     leave(nodeId) {
         if (this.nodes[nodeId] != null) {
             this.nodes[nodeId].forEach((neighbor) => this.nodes[neighbor].delete(nodeId))
-            delete this.nodes[nodeId]
         }
+        delete this.nodes[nodeId]
     }
 
     isEmpty() {
--- a/src/logic/Tracker.js
+++ b/src/logic/Tracker.js
@@ -92,6 +92,11 @@ module.exports = class Tracker extends EventEmitter {
     _updateNode(node, streams) {
         let newNode = true
 
+        if (streams === {}) {
+            this._removeNode(node)
+            return
+        }
+
         // Add or update
         Object.entries(streams).forEach(([streamKey, { inboundNodes, outboundNodes }]) => {
             if (this.overlayPerStream[streamKey] == null) {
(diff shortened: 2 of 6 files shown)


src/logic/OverlayTopology.js
```diff
@@ -34,13 +34,16 @@ class OverlayTopology {
     leave(nodeId) {
         if (this.nodes[nodeId] != null) {
             this.nodes[nodeId].forEach((neighbor) => this.nodes[neighbor].delete(nodeId))
-            delete this.nodes[nodeId]
         }
+        delete this.nodes[nodeId]
```
harbu: I'm unsure as to why this needs to be moved here? Do we have a case where `this.nodes[nodeId] === null`?

src/logic/Tracker.js
```diff
@@ -92,6 +92,11 @@ module.exports = class Tracker extends EventEmitter {
     _updateNode(node, streams) {
         let newNode = true
 
+        if (streams === {}) {
```
harbu, line 95: Is this `if`-block really needed? Doesn't the loop below under comment `// Remove` effectively achieve the same thing?

src/logic/Tracker.js
```diff
@@ -164,10 +170,21 @@ module.exports = class Tracker extends EventEmitter {
...
     }
 
+    _leaveAndCheckEmptyOverlay(streamKey, overlayTopology, node) {
+        overlayTopology.leave(node)
+
+        if (overlayTopology.isEmpty()) {
+            delete this.overlayPerStream[streamKey]
+        }
+
+        return overlayTopology
```
harbu: Why return?

test/integration/tracker.test.js
```diff
@@ -1,38 +1,112 @@
...
+        await subscriberOne.stop()
+        await subscriberTwo.stop()
+        await tracker.stop()
+    })
+
+    it('should be able to start tracker, two nodes, receive statuses, create overlayPerStream for streams, then stop them successfully', async () => {
+        expect(tracker.protocols.trackerServer.endpoint.connections.size).toBe(0)
+        expect(tracker.overlayPerStream).toEqual({})
+
+        await subscriberOne.addBootstrapTracker(tracker.getAddress())
```
harbu: Doesn't return promise, no need to await.

test/integration/tracker.test.js
```diff
@@ -1,38 +1,112 @@
...
-        await node2.addBootstrapTracker(tracker.getAddress())
+        expect(Object.keys(tracker.overlayPerStream)).toEqual(['stream-1::0', 'stream-2::2'])
+        expect(tracker.overlayPerStream['stream-1::0'].state()).toEqual({
+            subscriberOne: []
+        })
+        expect(tracker.overlayPerStream['stream-2::2'].state()).toEqual({
+            subscriberOne: []
+        })
 
+        await subscriberTwo.addBootstrapTracker(tracker.getAddress())
```
harbu: Doesn't return promise, no need to await.

test/integration/tracker.test.js
```diff
@@ -1,38 +1,112 @@
...
+            subscriberTwo: ['subscriberOne']
+        })
     })
 
-    afterAll(async () => {
-        await node1.stop()
-        await node2.stop()
-        await tracker.stop()
+    it('tracker should update correctly overlayPerStream on subscribe/unsubscribe', async () => {
+        await subscriberOne.addBootstrapTracker(tracker.getAddress())
```
harbu: Doesn't return promise, no need to await.

test/integration/tracker.test.js
```diff
@@ -1,38 +1,112 @@
...
+        await subscriberTwo.addBootstrapTracker(tracker.getAddress())
```
harbu: Doesn't return promise, no need to await.

test/unit/OverlayTopology.test.js
```diff
@@ -2,284 +2,287 @@ const OverlayTopology = require('../../src/logic/OverlayTopology')
 
 const maxNeighborsPerNodeArray = [4, 8, 12, 16]
 
-test('forming overlay topology', () => {
-    const topology = new OverlayTopology(3, (arr) => arr, (arr) => arr[0])
+describe('tracker creation', () => {
```
harbu: tracker creation?

test/integration/tracker.test.js
```diff
@@ -1,38 +1,112 @@
...
+            subscriberTwo: []
+        })
+
+        // console.log('subscriberOne.unsubscribeFromStream(s1)')
+        subscriberOne.unsubscribeFromStream(s1)
+
+        const res = {
+            subscriberTwo: []
+        }
+        await waitForCondition(() => JSON.stringify(tracker.overlayPerStream['stream-1::0'].state()) === JSON.stringify(res))
```
harbu: simpler condition

harbu (changes requested):

src/logic/Tracker.js
harbu: Discussed this and we decided to keep it as is. Invoking `_removeNode` directly is slightly more performant (we don't have to invoke the filter function on every element). Also this choice has the benefit of explicitly communicating what we do when we receive an empty object.

harbu (approved):
LGTM

PR #322, opened 2019-08-13
NET-47: UNSEEN_MESSAGE_RECEIVED event

harbu:
- Add new event `UNSEEN_MESSAGE_RECEIVED` to class `Node` that is emitted after duplicate detection but before actual message propagation. Also refactor some of the logic between duplicate detection and propagation a little.
- `NetworkNode#addMessageListener` now binds callbacks to event  `UNSEEN_MESSAGE_RECEIVED` instead of `MESSAGE_PROPAGATED`.
- Fix few tests

### DIFF
--- a/test/integration/message-propagation.test.js
+++ b/test/integration/message-propagation.test.js
@@ -143,6 +143,42 @@ describe('message propagation in network', () => {
         ])
         expect(n2Messages).toEqual(n1Messages)
         expect(n3Messages).toEqual(n2Messages)
-        expect(n4Messages).toEqual([])
+        expect(n4Messages).toEqual([
+            {
+                streamId: 'stream-2',
+                streamPartition: 0,
+                payload: {
+                    messageNo: 100
+                },
+            },
+            {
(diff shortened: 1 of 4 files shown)


test/integration/message-propagation.test.js
```diff
@@ -143,6 +143,42 @@ describe('message propagation in network', () => {
         ])
         expect(n2Messages).toEqual(n1Messages)
         expect(n3Messages).toEqual(n2Messages)
-        expect(n4Messages).toEqual([])
+        expect(n4Messages).toEqual([
```
harbu, line 146: Node `n4` is not connected to any other nodes in this test. Before this change set, message listener callbacks for this node were never invoked because `MESSAGE_PROPAGATED` event was never emitted. But now we bind to `UNSEEN_MESSAGE_RECEIVED` hence message handlers are invoked.

mirpo: 👍

mirpo (approved):

PR #331, opened 2019-09-20
Devops 121 docker dev

mthambipillai:
Notes:
- The Dockerfile is actually only for the tracker
- I had to remove the host param when creating the WebSocket server so that it would be accessible from outside the Docker container

### DIFF
--- a/.travis_scripts/deploy_docker.sh
+++ b/.travis_scripts/deploy_docker.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+docker login -u "${DOCKER_USER}" -p "${DOCKER_PASS}"
+if [ $1 = "staging" ]; then
+    docker push $OWNER/$IMAGE_NAME:$TAG
+elif [ $1 = "production" ]; then
+    docker build -t $OWNER/$IMAGE_NAME:$TAG .
+    #docker tag $OWNER/$IMAGE_NAME:$TAG $OWNER/$IMAGE_NAME:$TRAVIS_TAG
+    docker push $OWNER/$IMAGE_NAME:$TAG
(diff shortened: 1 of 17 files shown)


.travis_scripts/deploy_docker.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+docker login -u "${DOCKER_USER}" -p "${DOCKER_PASS}"
+if [ $1 = "staging" ]; then
+    docker push $OWNER/$IMAGE_NAME:$TAG
+elif [ $1 = "production" ]; then
+    docker build -t $OWNER/$IMAGE_NAME:$TAG .
+    #docker tag $OWNER/$IMAGE_NAME:$TAG $OWNER/$IMAGE_NAME:$TRAVIS_TAG
```
mirpo: Please remove commented lines.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Use official node runtime as base image
+FROM node:8.11.4-alpine
```
mirpo: We use node 10

bin/tracker.js
```diff
@@ -5,7 +5,7 @@ const util = require('util')
 const { startTracker } = require('../src/composition')
 
 const port = process.argv[2] || 30300
-const ip = process.argv[3] || '127.0.0.1'
+const ip = process.argv[3]
```
mirpo, line 8: We use it ./bin/ scripts, they might fail now, better run them to recheck

mirpo (changes requested):

bin/tracker.js
mthambipillai: Ran them locally, none fails

mirpo (approved):

mirpo:
LGFM!

PR #338, opened 2019-10-03
NET-60: cleanup README

harbu:


README.md
```diff
@@ -1,26 +1,34 @@
...
 - [Installation](#installation)
 - [Integration](#integration)
-- [Getting started](#getting-started)
-- [Advanced integrations](#advanced-integrations)
 - [Architectural decisions](https://github.com/streamr-dev/network/wiki)
-- [Component reference](#component-reference)
 - [Examples](#examples)
-- [Troubleshooting](#troubleshooting)
 - [Development](#development)
 - [Releasing](#releasing)
```
hpihkala, line 28: Can't comment on the correct line (38), but the `Screenshot of the demo running` URL is broken (404). Is the `network-runner` useful, maintained, and up to date? Or should we remove that liability by removing the Demo section from here?

README.md
```diff
@@ -1,26 +1,34 @@
 # Streamr network 
  ![Travis](https://travis-ci.com/streamr-dev/network.svg?token=qNNVCnYJo1fz18VTNpPZ&branch=master)
  
-> P2P network for the real-time data with storage support
+> Peer-to-peer publish-subscribe network for real-time data with support for long-term data persistence
+
+This repository/package contains an extendable server-side implementation of the 
+[Streamr protocol](https://github.com/streamr-dev/streamr-client-protocol-js) written in Node.js.
```
hpihkala: (voices in my head: do we want to rule out browsers? is node.js a language?)
```suggestion
[Streamr protocol](https://github.com/streamr-dev/streamr-client-protocol-js) in JS.
```

hpihkala: Would it be better to link to protocol spec instead of JS impl?
```suggestion
[Streamr protocol](https://github.com/streamr-dev/streamr-specs/blob/master/PROTOCOL.md) written in Node.js.
```

README.md
```diff
@@ -1,26 +1,34 @@
 # Streamr network 
  ![Travis](https://travis-ci.com/streamr-dev/network.svg?token=qNNVCnYJo1fz18VTNpPZ&branch=master)
  
-> P2P network for the real-time data with storage support
+> Peer-to-peer publish-subscribe network for real-time data with support for long-term data persistence
+
+This repository/package contains an extendable server-side implementation of the
```
hpihkala: Only server-side? I guess it could work for clients too? (ofc not tested in browsers yet though)

README.md
```diff
@@ -45,30 +53,10 @@ It is also possible to install Streamr Network globally (using `npm install @str
 
 Integration into existing project could be found in [Examples](./examples)
```
hpihkala: This line has some language problems, and it's unclear what "Integration into existing project" means. If the point is to point the reader to the Examples, I recommend changing the title to `Examples` and then rewriting the description what can be found there.

README.md
```diff
@@ -45,30 +53,10 @@ It is also possible to install Streamr Network globally (using `npm install @str
...
-
-## Advanced integrations
-
-TODO
-
-## Component reference
-
-TODO
-
 ## Examples
```
hpihkala, line 41: Oh alright, here we have the Examples title, so the above Integrations can maybe be completely removed (don't need two sections that both point to the `examples` folder). The content here is ultra-brief, needs expansion. What kind of examples can be found there? What/who are they useful for?

README.md
```diff
@@ -79,7 +67,7 @@ Run the tests:
 
     npm run test
 
-Run example of network (10 nodes):
+Run an example network locally (10 nodes):
```
hpihkala, line 56: Ah nice! If this is so easy, then the `network-runner` demo at the top is unnecessary?

hpihkala (changes requested):
A few thoughts while reading

README.md
harbu: Removed demo section

README.md
harbu: Fixed

README.md
harbu: A bit hard to define. It's not purely server but yeah not a "light client" either. Anyways, removed term "server-side".

README.md
harbu: Removed this paragraph.

README.md
harbu: Re-did writing of Examples section.

PR #339, opened 2019-10-03
Use streams

mirpo:
- replace reading and writing to websockets with stream
- replace closing socket with destroying stream, for preventing running not needed events

I don't how it will affect performance, because I use again websocket/ws library. To be tested in broker.

### DIFF
--- a/src/connection/WsEndpoint.js
+++ b/src/connection/WsEndpoint.js
@@ -8,7 +8,7 @@ const { EventEmitter } = require('events')
 const url = require('url')
 
 const createDebug = require('debug')
-const WebSocket = require('@streamr/sc-uws')
+const WebSocket = require('ws')
 
 const { disconnectionReasons } = require('../messages/messageTypes')
 const Metrics = require('../metrics')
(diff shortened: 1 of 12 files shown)


src/connection/WsEndpoint.js
```diff
@@ -339,7 +310,11 @@ class WsEndpoint extends EventEmitter {
             return
         }
 
-        ws.on('message', (message) => {
+        const duplex = WebSocket.createWebSocketStream(ws, {
```
harbu, line 313: Should we have `websocket-stream` library in package.json?

harbu (changes requested):
LGTM; see one comment about package.json

src/connection/WsEndpoint.js
mirpo: It's not a websocket-stream its a function from ws https://github.com/websockets/ws#use-the-nodejs-streams-api

harbu: waaat.. nice

harbu (approved):

PR #349, opened 2019-10-30
Commander

mirpo:
- replace process.argv with commander
- remove not needed bin files
- update libs


bin/publisher.js
```diff
@@ -2,27 +2,37 @@
...
-const intervalInMs = process.argv[6] || 200
+program
+    .version(CURRENT_VERSION)
+    .option('--port <port>', 'port', 30302)
+    .option('--ip <ip>', 'ip', '127.0.0.1')
+    .option('--trackers <trackers>', 'trackers', (value) => value.split(','), ['ws://127.0.0.1:30300'])
+    .option('--streamId <streamId>', 'streamId to publish', 'default-stream-id')
+    .option('--metrics <metrics>', 'log metrics', false)
+    .option('--intervalInMs <intervalInMs>', 'interval to publish in ms', 200)
+    .description('Run subscriber')
```
harbu: You mean `run publisher`?

harbu (changes requested):
Looks much better, proper argument handling with Commander. 🎖 

Dockerfile needs to be updated though.

harbu (approved):

PR #350, opened 2019-10-31
Fix Dockerfile & tracker.js

harbu:
Current combination of Dockerfile and tracker.js would bind on 127.0.0.1 causing it to bind on interface loopback `lo`. Fixed it so that default host is null.


mirpo (approved):
lgfm

PR #354, opened 2019-11-01
Implement singleton of BasicProtocol

mirpo:


test/integration/l3-resend-requests.test.js
```diff
@@ -111,7 +111,7 @@ describe('resend requests are fulfilled at L3', () => {
         await storageNode.stop()
     })
 
-    beforeEach(() => {
+    beforeEach(async () => {
```
harbu: Why add `async`?

test/integration/unsubscribe-from-stream.test.js
```diff
@@ -20,16 +21,11 @@ describe('node unsubscribing from a stream', () => {
...
-        await wait(500)
-
         nodeA.subscribe('s', 2)
         nodeB.subscribe('s', 2)
 
-        await wait(500)
-
         await waitForEvent(nodeB, Node.events.NODE_SUBSCRIBED)
-        await waitForEvent(nodeA, Node.events.NODE_SUBSCRIBED)
+        await waitForEvent(tracker.protocols.trackerServer, TrackerServer.events.NODE_STATUS_RECEIVED)
```
harbu: Not sure I understand this change. Should we have a think thru the `await`s here?

harbu (changes requested):
LGTM; few points

harbu (approved):

PR #361, opened 2019-11-18
prepare for streamr-client-protocol-js 3.0.0

harbu:
Prepare for streamr-client-protocol-js major version upgrade 3.0.0 by using pre-release version 3.0.0-0 and fixing issues.


mirpo (approved):
🏅

PR #366, opened 2019-11-21
Cancel resend when node disconnects

### LINKED ISSUE #358: Cancel resend when requesting client drops
Currently resends are not apparently canceled when the requesting client disconnects.

harbu:
Fixes #358 

If node X has ongoing resend request(s) from node Y, and then node X suddenly decides to disconnect, the ongoing resend requests should be canceled.

- [x] Integration test reproducing issue
- [x] First go at solution
- [x] Refactoring


mirpo (commented):
🥇

mirpo (approved):

PR #374, opened 2019-12-04
Udp libs, add benchmark for the tracker sendInstructions, install streamr-client-protocol v4.0.1, replace subId with requestId

mirpo:


src/logic/Node.js
```diff
@@ -163,13 +163,13 @@ class Node extends EventEmitter {
...
             } catch (e) {
-                this.debug('failed to connect to node at %s (%j)', nodeAddress, e)
+                this.debug('failed to connect to node at %s (%j), to subscribe streamId %s', nodeAddress, e, streamId)
                 return
             }
             try {
                 await this._subscribeToStreamOnNode(node, streamId)
             } catch (e) {
-                this.debug('failed to subscribe to node %s (%s)', node, e)
+                this.debug('failed to subscribe to node %s (%s), streamId %s', node, e, streamId)
```
harbu: Should error be printed with `%j` like above?

src/logic/Tracker.js
```diff
@@ -118,7 +118,7 @@ module.exports = class Tracker extends EventEmitter {
             .filter(([streamKey, _]) => !currentStreamKeys.has(streamKey))
             .forEach(([streamKey, overlayTopology]) => this._leaveAndCheckEmptyOverlay(streamKey, overlayTopology, node))
 
-        if (!newNode) {
+        if (newNode) {
```
harbu, line 121: Hmm.. the existing code looks like a bug. We (re)assign `newNode` on each iteration of `forEach` to either truthy or false. But the final value depends on the last item of  `Object.entries(streams)`, which is quite arbitrary?

harbu (approved):
LGTM; I found a slight weirdness with existing code but nothing preventing this from being landed.

PR #383, opened 2020-01-23
use StreamMessage returned from Storage

mthambipillai:
Solves this bug: https://streamr.atlassian.net/browse/NET-99
Requires to change the Storage: https://github.com/streamr-dev/broker/pull/118


harbu (approved):
LGTM

PR #385, opened 2020-01-31
WebRTC

harbu:


src/connection/WebRtcEndpoint.js
```diff
@@ -0,0 +1,113 @@
...
+            const connection = this.connections[originatorId]
+            if (connection) {
+                const description = new RTCSessionDescription(answer)
+                await connection.setRemoteDescription(description)
+            }
+        })
+
+        rtcSignaller.setOfferListener(async (originatorId, offer) => {
+            const connection = this.connections[originatorId]
+            if (connection) {
```
ptesavol, line 105: This seems to only work if the receiver of the offer has already called connect(), which is not always the case. I think it would be best to move the creation of the RTCPeerConnection to a helper function that is called from both connect() and from here if connect() has not been called yet.

src/connection/WebRtcEndpoint.js
```diff
@@ -0,0 +1,113 @@
...
+        connection.onsignalingstatechange = (event) => {
+            console.log('onsignalingstatechange', this.id, connection.connectionState, event)
+        }
+        connection.oniceconnectionstatechange = (event) => {
+            console.log('oniceconnectionstatechange', this.id, event)
+        }
+        connection.onicegatheringstatechange = (event) => {
+            console.log('onicegatheringstatechange', this.id, event)
+        }
+        connection.onnegotiationneeded = (event) => {
```
ptesavol: It is best to start creating the offer in this onnegotiationneeded callback. This gets fired whenever new data channel is added to the RTCPeerConnection and when network connection settings change in some way or another (ie. reconnect after short disconnection)

src/connection/WebRtcEndpoint.js
```diff
@@ -0,0 +1,113 @@
...
+            console.log('onnegotiationneeded', this.id, event)
+        }
+
+        const isOffering = this.id < targetPeerId
+        if (isOffering) {
+            this._initDataChannel(targetPeerId, connection.createDataChannel())
```
ptesavol: here you can set whether you want the dataChannel to be reliable of not, and give a name to the data channel like this: 

let channel = connection.createDataChannel("jsonrpcchannel", { reliable: true });

it is also a good idea to set the binary type to arraybuffer here, in case we will transmit binary data at some point:

channel.binaryType = "arraybuffer";

harbu: Looks like the data channel negotiation can be mode symmetrical in the code by using `negotiated: true` option. Then both ends just call `createDataChannel` see: https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createDataChannel#Examples

I wonder if there is any downsides to enabling this?

harbu: I guess both parties need to know the `channelId` beforehand for this approach to work. If we just use one channel to transmit data this should be easy. If, however, we wanted to use a channel per stream approach we need negotiation.

src/connection/WebRtcEndpoint.js
harbu: Good point. I'll make a test for this case and then fix it to work when the receiver has not yet invoked `connect`.

src/connection/WebRtcEndpoint.js
harbu: done

src/connection/WebRtcEndpoint.js
```diff
@@ -0,0 +1,131 @@
...
+
+        const configuration = {
+            iceServers: this.stunUrls.map((url) => ({
+                urls: url
+            }))
+        }
+        const connection = new RTCPeerConnection(configuration)
+        const dataChannel = connection.createDataChannel('streamrDataChannel', {
+            id: 0,
+            negotiated: true
```
mirpo, line 272: Did you discuss other flags?
```
ordered: If the data channel should guarantee order or not
maxPacketLifeTime: The maximum time to try and retransmit a failed message
maxRetransmits: The maximum number of times to try and retransmit a failed message
```

src/connection/WebRtcEndpoint.js
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        const configuration = {
+            iceServers: this.stunUrls.map((url) => ({
+                urls: url
+            }))
+        }
+        const connection = new RTCPeerConnection(configuration)
+        const dataChannel = connection.createDataChannel('streamrDataChannel', {
+            id: 0,
```
mirpo, line 271: why zero and not some `uuid`?

src/connection/WebRtcEndpoint.js
```diff
@@ -0,0 +1,131 @@
...
+            negotiated: true
+        })
+
+        this.connections[targetPeerId] = connection
+        this.dataChannels[targetPeerId] = dataChannel
+
+        const isOffering = this.id < targetPeerId
```
mirpo: why do we need this isOffering?

src/connection/WebRtcEndpoint.js
```diff
@@ -0,0 +1,131 @@
...
+                this.rtcSignaller.offer(routerId, targetPeerId, offer)
+            }
+        }
+        connection.onicecandidate = (event) => {
+            if (event.candidate != null) {
+                this.rtcSignaller.onNewIceCandidate(routerId, targetPeerId, event.candidate)
+            }
+        }
+        connection.onconnectionstatechange = (event) => {
+            console.log('onconnectionstatechange', this.id, connection.connectionState, event)
```
mirpo: `connectionState` 🥇

src/connection/WebRtcEndpoint.js
```diff
@@ -0,0 +1,131 @@
...
+            console.log('onicegatheringstatechange', this.id, event)
+        }
+        dataChannel.onopen = (event) => {
+            this.emit(events.PEER_CONNECTED, targetPeerId)
+        }
+        dataChannel.onclose = (event) => {
+            this.emit(events.PEER_DISCONNECTED, targetPeerId)
+        }
+        dataChannel.onerror = (event) => {
+            console.error(event)
```
mirpo, line 336: maybe we need here `this.emit(events.PEER_DISCONNECTED, targetPeerId)`

mirpo: And I don't see where you remove connections from `this.connections` and do some `.close()`

src/logic/RtcSignaller.js
```diff
@@ -0,0 +1,59 @@
...
+    answer(routerId, targetPeerId, answer) {
+        this.trackerNode.sendRtcAnswer(routerId, targetPeerId, this.id, answer)
+    }
+
+    onNewIceCandidate(routerId, targetPeerId, candidate) {
+        this.trackerNode.sendIceCandidate(routerId, targetPeerId, this.id, candidate)
+    }
+
+    setOfferListener(fn) {
+        this.offerListener = fn
```
mirpo, line 57: `fn` - arghhh

src/protocol/TrackerNode.js
```diff
@@ -30,6 +34,21 @@ class TrackerNode extends EventEmitter {
         return this.basicProtocol.endpoint.send(trackerAddress, encoder.findStorageNodesMessage(streamId))
     }
 
+    sendRtcOffer(trackerId, targetNode, originatorNode, data) {
```
mirpo: I think it's better to remove `originatorNode` from parameters, it's just internal info from endpoint

test/unit/WebRtcEndpoint.test.js
```diff
@@ -0,0 +1,127 @@
...
+            'streamr-peer-id': 'node-2',
+            'streamr-peer-type': peerTypes.NODE
+        }, null)))
+
+        trackerNode1.connectToTracker(tracker.getAddress())
+        await waitForEvent(tracker.protocols.trackerServer, TrackerServer.events.NODE_CONNECTED)
+        trackerNode2.connectToTracker(tracker.getAddress())
+        await waitForEvent(tracker.protocols.trackerServer, TrackerServer.events.NODE_CONNECTED)
+
+        endpoint1 = new WebRtcEndpoint('node-1', ['stun:stun.l.google.com:19302'], new RtcSignaller('node-1', trackerNode1))
```
mirpo: later we'll need smth like https://www.npmjs.com/package/stunsrv

test/unit/WebRtcEndpoint.test.js
```diff
@@ -0,0 +1,127 @@
...
+                hello: 'world'
+            }))
+        }
+
+        for (let i = 0; i < 10; ++i) {
+            setTimeout(sendFrom1To2, 10 * i)
+            setTimeout(sendFrom2To1, 10 * i + 5)
+        }
+
+        await waitForCondition(() => ep1NumOfReceivedMessages > 9)
```
mirpo, line 81: ep1NumOfReceivedMessages == 10 ?

mirpo (changes requested):
LGFM, commander!
What we are going to do with WS endpoint? remove it at all? or keep?

src/connection/WebRtcEndpoint.js
harbu: Good thing about WebRTC is that these are quite easy to change in future. I'd go with defaults now unless we have a reason to change something.

src/connection/WebRtcEndpoint.js
harbu: Both sides must agree on id beforehand when using "out-of-band" data channels. Stackoverflow answer here https://stackoverflow.com/questions/43788872/how-are-data-channels-negotiated-between-two-peers-with-webrtc provides more insight into the two different modes of data channel setup. Basically for a fixed amount of data channels (we have one now) it is better to use out-of-band negotiation.

src/connection/WebRtcEndpoint.js
harbu: WebRTC doesn't work if both sides try to send offer to each other at same time. Because `tracker` currently sends instruction to both participants to connect to each other, I added this check to ensure that only one side sends offer and the other answers. I could actually add a comment here to explain this. Later on we may want to change tracker to send connection instruction to only one participant, perhaps, needs further discussion.

src/connection/WebRtcEndpoint.js
harbu: Hopefully not as problematic as readyState...

test/unit/WebRtcEndpoint.test.js
harbu: fixed

src/connection/WebRtcEndpoint.js
```diff
@@ -0,0 +1,274 @@
...
+                    this.messageQueue[targetPeerId].pop()
+                    queueItem.delivered()
+                } catch (e) {
+                    console.error(e)
+                    queueItem.incrementTries()
+                    if (!queueItem.isFailed() && this.flushTimeOutRefs[targetPeerId] == null) {
+                        this.flushTimeOutRefs[targetPeerId] = setTimeout(() => {
+                            delete this.flushTimeOutRefs[targetPeerId]
+                            this._attemptToFlushMessages(targetPeerId)
+                        }, 100)
```
mirpo, line 193: real time? :)

src/logic/Node.js
```diff
@@ -152,33 +154,28 @@ class Node extends EventEmitter {
...
+        const tracker = instructionMessage.getSource()
+        const successfulNodeIds = []
 
         this.debug('received instructions for %s', streamId)
         this.subscribeToStreamIfHaveNotYet(streamId)
 
-        await Promise.all(nodeAddresses.map(async (nodeAddress) => {
-            let node
+        await Promise.all(assignedNodeIds.map(async (nodeId) => {
+            await this.protocols.nodeToNode.connectToNode(nodeId, tracker)
```
mirpo: you need to update all this logic from master

test/integration/tracker-node-reconnect-instructions.test.js
```diff
@@ -52,7 +52,6 @@ describe('Check tracker instructions to node', () => {
         let secondCheck = false
 
         otherNodes[1].protocols.nodeToNode.endpoint.once(endpointEvents.PEER_DISCONNECTED, (peerId, reason) => {
-            expect(reason).toBe(disconnectionReasons.NO_SHARED_STREAMS)
             firstCheck = true
```
mirpo: why you removed `NO_SHARED_STREAMS` ?

test/integration/unsubscribe-from-stream.test.js
```diff
@@ -77,6 +79,7 @@ describe('node unsubscribing from a stream', () => {
     })
 
     test('connection between nodes is not kept if no shared streams', async () => {
+        await wait(100)
         nodeB.unsubscribe('s', 2)
```
mirpo, line 72: maybe we can await for some condition?

mirpo (approved):
LGFM

src/connection/WebRtcEndpoint.js
```diff
@@ -0,0 +1,295 @@
...
+                this.rtcSignaller.onNewIceCandidate(routerId, targetPeerId, event.candidate)
+            }
+        }
+        connection.onconnectionstatechange = (event) => {
+            this.debug('onconnectionstatechange', this.id, targetPeerId, connection.connectionState, event)
+        }
+        connection.onsignalingstatechange = (event) => {
+            this.debug('onsignalingstatechange', this.id, targetPeerId, connection.connectionState, event)
+        }
+        connection.oniceconnectionstatechange = (event) => {
```
juslesan, line 316: If connection is set to failed it seems that it is never repaired, causing the network to break. It seems that in some cases the other end of the connection will disconnect if the state is set to failed, but the connection is never re-established.

harbu:
Closed. We went with `node-datachannel` library instead of `wrtc`.

PR #387, opened 2020-02-14
Fix ws endpoint crashing

mirpo:


juslesan (approved):

test/unit/tracker.test.js
```diff
@@ -1,10 +1,55 @@
...
+
+    afterEach(async () => {
         await tracker.stop()
     })
+
+    it('tracker should be able to start and stop successfully', async () => {
+        expect(tracker.getAddress()).toEqual('ws://127.0.0.1:30300')
+    })
+
+    it('tracker must check all required information for new incoming connection and not crash', async () => {
```
harbu: This is testing functionality of `WsEndpoint` right? Should these tests be in 'ws-endpoint.test.js' or some new test file dedicated to WsEndpoint?

harbu (changes requested):
LGTM; but I think the tests should be moved to a different place

test/unit/tracker.test.js
mirpo: Soon only tracker will use websocket

harbu (approved):

PR #391, opened 2020-02-24
Fix promises

mirpo:
- Installed `eslint-plugin-promise`, so it'll force some promise rules.
- Change a bit `_subscribeToStreamOnNode`, so no more `uncaught exception in promise` from `connect` function
- @harbu need your help. in file: `test/unit/ResendHandler.test.js` I found missed `await` or it was expected? Anyway when I put `await` back, `streamHasEnded` is always `true`.

### DIFF
--- a/src/logic/Node.js
+++ b/src/logic/Node.js
@@ -347,8 +347,6 @@ class Node extends EventEmitter {
                 })
             })
         }
-
-        return false
     }
 
     async _unsubscribeFromStreamOnNode(node, streamId) {
--- a/src/logic/Node.js
+++ b/src/logic/Node.js
@@ -345,7 +345,6 @@ class Node extends EventEmitter {
                     streamId,
                     node
                 })
-                return true
             })
         }
 
--- a/src/logic/Node.js
+++ b/src/logic/Node.js
@@ -333,6 +333,7 @@ class Node extends EventEmitter {
         }
     }
 
+    // eslint-disable-next-line consistent-return
     async _subscribeToStreamOnNode(node, streamId) {
         if (!this.streams.hasInboundNode(streamId, node)) {
             // more strict, so when we get reject from connect, it will be caught
--- a/src/logic/Node.js
+++ b/src/logic/Node.js
@@ -335,17 +335,21 @@ class Node extends EventEmitter {
 
     async _subscribeToStreamOnNode(node, streamId) {
         if (!this.streams.hasInboundNode(streamId, node)) {
-            await this.protocols.nodeToNode.sendSubscribe(node, streamId)
+            // more strict, so when we get reject from from connect, it will be catched
+            return this.protocols.nodeToNode.sendSubscribe(node, streamId).then(() => {
+                this.streams.addInboundNode(streamId, node)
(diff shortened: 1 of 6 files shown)


src/logic/resendStrategies.js
```diff
@@ -176,8 +176,11 @@ class ProxiedResend {
         this.nodeToNode.send(neighborId, this.request).then(() => {
             this.currentNeighbor = neighborId
             this._resetTimeout()
+            return true
```
harbu: I don't understand why we need to return boolean here, is it used somewhere?

src/logic/Node.js
```diff
@@ -335,17 +335,21 @@ class Node extends EventEmitter {
...
+                // TODO get prove message from node that we successfully subscribed
+                this.emit(events.NODE_SUBSCRIBED, {
+                    streamId,
+                    node
+                })
+                return true
             })
         }
+
+        return false
```
harbu: Don't understand need for `return false` and `return true`? Are they used somewhere?

test/unit/ResendHandler.test.js
```diff
@@ -375,12 +375,13 @@ describe('ResendHandler', () => {
             }
 
             let streamHasEnded = false
-            waitForStreamToEnd(resendHandler.handleRequest(request, 'source'))
+            await waitForStreamToEnd(resendHandler.handleRequest(request, 'source'))
```
harbu: No need for `await` since we use `promise.finally()` here that waits for promise to resolve/reject to call the callback. Actually I think we should not have `await` here for the test to be correct.

src/logic/Node.js
```diff
@@ -335,17 +335,21 @@ class Node extends EventEmitter {
...
-
-            this.streams.addInboundNode(streamId, node)
-            this.streams.addOutboundNode(streamId, node)
-
-            // TODO get prove message from node that we successfully subscribed
-            this.emit(events.NODE_SUBSCRIBED, {
-                streamId,
-                node
+            // more strict, so when we get reject from connect, it will be caught
+            return this.protocols.nodeToNode.sendSubscribe(node, streamId).then(() => {
```
harbu, line 340: To me the code looks equivalent to before. But we get better error messages with this new version?

harbu (changes requested):
Naice. 😎Added a few questions. Only thing I think should not be changed is the test.

src/logic/resendStrategies.js
mirpo: Add only to make happy eslint https://github.com/xjamundx/eslint-plugin-promise/blob/HEAD/docs/rules/always-return.md
Anyway, I think we'll add later strict rules (https://github.com/streamr-dev/eslint-config-streamr/pull/9), so it'll catch all possible errors. For now => reverted back.

src/logic/Node.js
mirpo: Better messages and better code!

src/logic/resendStrategies.js
harbu: ok makes sense

harbu (approved):

PR #392, opened 2020-02-27
Update fork of uws with latest uWebsocket.js

mirpo:


juslesan (approved):

package.json
```diff
@@ -45,18 +45,20 @@
...
+    "qs": "^6.9.1",
     "speedometer": "^1.1.0",
-    "streamr-client": "^3.1.2",
-    "streamr-client-protocol": "^4.1.0",
-    "uuid": "^3.4.0"
+    "streamr-client": "^3.1.3",
+    "streamr-client-protocol": "^4.1.1",
+    "uWebSockets.js": "github:uNetworking/uWebSockets.js#v17.1.0",
+    "uuid": "^7.0.1",
+    "ws": "^7.2.1"
```
harbu: Should we remove `ws` from `devDependencies`?

src/connection/WsEndpoint.js
```diff
@@ -5,23 +5,49 @@ const events = Object.freeze({
...
-    Object.entries(o).forEach(([k, v]) => {
-        transformedO[k.toLowerCase()] = v
-    })
-    return transformedO
+const ab2str = (buf) => Buffer.from(buf).toString('utf8')
+
+// TODO uWS will soon rename end -> close and end -> terminate
+// https://github.com/uNetworking/uWebSockets.js/blob/master/src/WebSocketWrapper.h#L66
+const closeWs = (ws, code, reason) => {
+    if (ws.terminate !== undefined) {
```
harbu, line 25: Should this be `ws.close !== undefined` ?

src/connection/WsEndpoint.js
```diff
@@ -76,24 +101,41 @@ class WsEndpoint extends EventEmitter {
...
-
-        this.wss.verifyClient = (info) => {
-            const parameters = url.parse(info.req.url, true)
-            const { address } = parameters.query
-
-            if (this.isConnected(address)) {
-                this.debug('already connected to %s, readyState %d', address, this.connections.get(address).readyState)
-                this.debug('closing existing socket')
-                this.connections.get(address).close()
+        this.wss.get('/', (res, req) => {
```
harbu, line 101: Do I understand correctly that an additional HTTP forwarding step is involved when connecting to a node?

src/connection/WsEndpoint.js
```diff
@@ -76,24 +101,41 @@ class WsEndpoint extends EventEmitter {
...
+            drain: (ws) => {
+                this.debug(`WebSocket backpressure: ${ws.getBufferedAmount()}`)
+            },
+            close: (ws, code, message) => {
+                const reason = ab2str(message)
+
+                const connection = this.connections.get(ws.address)
+
+                if (connection) {
+                    this.emit('close', ws, code, reason) // event for back compatibility
```
harbu: What is this `close` event? Is the intention to emit this event on the `wss` object? If yes, then I think it should be `this.wss.emit`.

src/connection/WsEndpoint.js
```diff
@@ -206,11 +248,11 @@ class WsEndpoint extends EventEmitter {
 
     onReceive(peerInfo, address, message) {
         this.metrics.inc('onReceive')
-        this.debug('received from %s [%s] message "%s"', peerInfo, address, message)
-        this.emit(events.MESSAGE_RECEIVED, peerInfo, message)
+        this.debug('<=== received from %s [%s] message "%s"', peerInfo, address, message)
+        setImmediate(() => this.emit(events.MESSAGE_RECEIVED, peerInfo, message))
```
harbu: Why the addition of `setImmediate`?

src/connection/WsEndpoint.js
```diff
@@ -247,21 +289,38 @@ class WsEndpoint extends EventEmitter {
...
+                        headers: toHeaders(this.peerInfo)
+                    }
+                )
 
-                ws.on('upgrade', (peerId, peerType) => {
-                    serverPeerInfo = new PeerInfo(peerId, peerType)
+                // catching headers
+                // eslint-disable-next-line no-underscore-dangle
+                ws._req.on('response', (res) => {
+                    const peerId = res.headers['streamr-peer-id']
```
harbu, line 304: Case sensitivity?

src/connection/WsEndpoint.js
```diff
@@ -247,21 +289,38 @@ class WsEndpoint extends EventEmitter {
...
+                        serverPeerInfo = new PeerInfo(peerId, peerType)
+                    }
                 })
 
                 ws.once('open', () => {
                     if (!serverPeerInfo) {
-                        ws.terminate()
+                        terminateWs(ws)
                         this.metrics.inc('connect:dropping-upgrade-never-received')
                         reject(new Error('dropping outgoing connection because upgrade event never received'))
```
harbu: I'm guessing we should use some other term than upgrade here so we don't confuse with the upgrade mechanism of HTTP -> Socket of WebSocket?

src/connection/WsEndpoint.js
```diff
@@ -292,14 +351,28 @@ class WsEndpoint extends EventEmitter {
...
         clearInterval(this.checkConnectionsInterval)
-        this.connections.forEach((connection) => {
-            connection.close(disconnectionCodes.GRACEFUL_SHUTDOWN, disconnectionReasons.GRACEFUL_SHUTDOWN)
-        })
 
         return new Promise((resolve, reject) => {
-            // uws has setTimeout(cb, 20000); in close event
-            this.wss.close()
-            resolve()
+            try {
```
harbu, line 351: Out of curiosity: why Promise instead of async/await?

src/connection/WsEndpoint.js
```diff
@@ -325,71 +397,98 @@ class WsEndpoint extends EventEmitter {
...
+            // see node_modules/uWebSockets.js/index.d.ts WebSocket definition
+            // eslint-disable-next-line no-param-reassign
+            ws.peerInfo = clientPeerInfo
+            // eslint-disable-next-line no-param-reassign
+            ws.address = address
 
-            this.debug('%s connected to me', address)
-            this._onNewConnection(ws, address, clientPeerInfo)
+            this.debug('<=== %s connecting to me', address)
+            this.emit('connection', ws) // event for back compatibility
```
harbu: What is this `connection` event? Is the intention to emit this event on the `wss` object? If yes, then I think it should be `this.wss.emit`.

src/connection/WsEndpoint.js
```diff
@@ -325,71 +397,98 @@ class WsEndpoint extends EventEmitter {
...
+            this._onNewConnection(ws, address, clientPeerInfo, false)
         } catch (e) {
-            this.debug('dropped incoming connection from %s because of %s', req.connection.remoteAddress, e)
-            this.metrics.inc('_onIncomingConnection:closed:missing-required-parameter')
-            ws.close(disconnectionCodes.MISSING_REQUIRED_PARAMETER, `${e}`)
+            this.debug('dropped incoming connection because of %s', e)
+            this.metrics.inc('_onIncomingConnection:closed:no-required-parameter')
```
harbu, line 426: missing is a more accurate word here since some of the parameters may be present

src/connection/WsEndpoint.js
```diff
@@ -325,71 +397,98 @@ class WsEndpoint extends EventEmitter {
...
+            closeWs(ws, 1002, e.toString())
         }
     }
 
-    _onNewConnection(ws, address, peerInfo) {
+    _onNewConnection(ws, address, peerInfo, out = true) {
```
harbu: Do we need parameters `address` and `peerInfo` since they are already found within "user data" of `ws`? (I.e. `ws.peerInfo` and `ws.address`)

src/connection/WsEndpoint.js
```diff
@@ -325,71 +397,98 @@ class WsEndpoint extends EventEmitter {
...
         // Handle scenario where two peers have opened a socket to each other at the same time.
         // Second condition is a tiebreaker to avoid both peers of simultaneously disconnecting their socket,
-        // thereby leaving no connection behind.
+        // thereby leaving no connection behind.g
```
harbu: typo

src/connection/WsEndpoint.js
```diff
@@ -325,71 +397,98 @@ class WsEndpoint extends EventEmitter {
...
+            this.debug('<=== %s connecting to me', address)
+            this.emit('connection', ws) // event for back compatibility
+            this._onNewConnection(ws, address, clientPeerInfo, false)
         } catch (e) {
-            this.debug('dropped incoming connection from %s because of %s', req.connection.remoteAddress, e)
-            this.metrics.inc('_onIncomingConnection:closed:missing-required-parameter')
-            ws.close(disconnectionCodes.MISSING_REQUIRED_PARAMETER, `${e}`)
+            this.debug('dropped incoming connection because of %s', e)
+            this.metrics.inc('_onIncomingConnection:closed:no-required-parameter')
+            closeWs(ws, 1002, e.toString())
```
harbu: Removed `disconnectionCodes.MISSING_REQUIRED_PARAMETER` on purpose?

src/connection/WsEndpoint.js
```diff
@@ -325,71 +397,98 @@ class WsEndpoint extends EventEmitter {
...
         // Handle scenario where two peers have opened a socket to each other at the same time.
         // Second condition is a tiebreaker to avoid both peers of simultaneously disconnecting their socket,
-        // thereby leaving no connection behind.
+        // thereby leaving no connection behind.g
         if (this.isConnected(address) && this.getAddress().localeCompare(address) === 1) {
-            this.metrics.inc('_onNewConnection:closed:dublicate')
+            this.metrics.inc('_onNewConnection:closed:duplicate')
             this.debug('dropped new connection with %s because an existing connection already exists', address)
-            ws.close(disconnectionCodes.DUPLICATE_SOCKET, disconnectionReasons.DUPLICATE_SOCKET)
+            closeWs(ws, 1000, disconnectionReasons.DUPLICATE_SOCKET)
```
harbu: `disconnectionCodes.DUPLICATE_SOCKET`?

src/connection/WsEndpoint.js
harbu: I don't think boolean parameter `out` has a "natural" default value that one could expect it to take without having to read the method signature. Thus I'd prefer we supply the value of the argument every time we call this function.

src/connection/WsEndpoint.js
```diff
@@ -325,71 +397,98 @@ class WsEndpoint extends EventEmitter {
...
         }
 
+        this.peerBook.add(address, peerInfo)
+        this.connections.set(address, ws)
+        this.metrics.set('connections', this.connections.size)
+        this.debug('added %s [%s] to connection list', peerInfo, address)
+        this.debug('%s connected to %s', out ? '===>' : '<===', address)
+        this.emit(events.PEER_CONNECTED, peerInfo)
+
+        return peerInfo
```
harbu: Maybe just `return true` if we only need return value for checking for success.

src/connection/WsEndpoint.js
```diff
@@ -405,29 +504,23 @@ class WsEndpoint extends EventEmitter {
...
-            reject(err)
-        })
+        // TODO add SSL support uWS.SSLApp()
+        const server = uWS.App()
 
-        wss.on('listening', () => {
-            resolve(wss)
+        server.listen(host, port, (token) => {
+            if (token) {
+                resolve([server, token])
```
harbu: Should `token` be re-named `listenSocket`? Token sounds like some string/number.

src/logic/Node.js
```diff
@@ -158,31 +158,32 @@ class Node extends EventEmitter {
...
-
-        this._sendStatusToAllTrackers()
-
-        const currentNodes = this.streams.isSetUp(streamId) ? this.streams.getAllNodesForStream(streamId) : []
-        const nodesToUnsubscribeFrom = currentNodes.filter((node) => !nodeIds.includes(node))
-
-        nodesToUnsubscribeFrom.forEach((node) => {
-            this._unsubscribeFromStreamOnNode(node, streamId)
-        })
+        if (nodeAddresses.length) {
```
harbu: If tracker sends us an empty list of nodeAddresses it means we should unsubscribe from all nodes. Thus I don't think it is correct to skip.

src/logic/Node.js
```diff
@@ -294,24 +295,10 @@ class Node extends EventEmitter {
         this.debug('stopping')
         this.resendHandler.stop()
         this._clearConnectToBootstrapTrackersInterval()
-        this._disconnectFromAllNodes()
```
harbu, line 315: Why are these removed? Shouldn't we disconnect when we stop Node?

src/logic/Tracker.js
```diff
@@ -162,10 +162,10 @@ module.exports = class Tracker extends EventEmitter {
         this.metrics.inc('_formAndSendInstructions')
         Object.keys(streams).forEach((streamKey) => {
             const instructions = this.overlayPerStream[streamKey].formInstructions(node)
-            Object.entries(instructions).forEach(async ([nodeId, newNeighbors]) => {
+            Object.entries(instructions).forEach(([nodeId, newNeighbors]) => {
                 try {
                     this.metrics.inc('sendInstruction')
-                    await this.protocols.trackerServer.sendInstruction(nodeId, StreamIdAndPartition.fromKey(streamKey), newNeighbors)
+                    this.protocols.trackerServer.sendInstruction(nodeId, StreamIdAndPartition.fromKey(streamKey), newNeighbors)
```
harbu: Some background info on why this change?

src/logic/Tracker.js
```diff
@@ -162,10 +162,10 @@ module.exports = class Tracker extends EventEmitter {
...
                     this.debug('sent instruction %j for stream %s to node %s', newNeighbors, streamKey, nodeId)
                 } catch (e) {
```
harbu: Dunno if this catch is useful anymore since an error will not be thrown on failure of delivery.

src/logic/Tracker.js
```diff
@@ -192,10 +192,10 @@ module.exports = class Tracker extends EventEmitter {
...
 
                 if (streamsToSubscribe.length) {
-                    streamsToSubscribe.forEach(async (streamKey) => {
+                    streamsToSubscribe.forEach((streamKey) => {
                         try {
                             this.metrics.inc('sendInstructionStorages')
-                            await this.protocols.trackerServer.sendInstruction(storageNode, StreamIdAndPartition.fromKey(streamKey), [])
+                            this.protocols.trackerServer.sendInstruction(storageNode, StreamIdAndPartition.fromKey(streamKey), [])
                             this.debug('sent instruction %j for stream %s to storage node %s', [], streamKey, storageNode)
                         } catch (e) {
```
harbu: Again, this catch is not gonna be really invoked  ever with `sendSync`.

src/protocol/TrackerServer.js
```diff
@@ -21,7 +21,7 @@ class TrackerServer extends EventEmitter {
 
     sendInstruction(receiverNodeId, streamId, listOfNodeIds) {
         const listOfNodeAddresses = listOfNodeIds.map((nodeId) => this.endpoint.resolveAddress(nodeId))
-        return this.endpoint.send(receiverNodeId, encoder.instructionMessage(streamId, listOfNodeAddresses))
+        this.endpoint.sendSync(receiverNodeId, encoder.instructionMessage(streamId, listOfNodeAddresses))
```
harbu: Btw what is difference between `sendSync` and `send`. Can we just move to use one of them?

test/integration/duplicate-connections-are-closed.test.js
```diff
@@ -1,22 +1,22 @@
...
-    let wss1
-    let ws1
-    let wss2
-    let ws2
+    const port1 = 28501
+    const port2 = 28502
+
     let wsEndpoint1
     let wsEndpoint2
+    let ws1
```
harbu: Unused vars

test/integration/duplicate-connections-are-closed.test.js
```diff
@@ -41,20 +41,21 @@ describe('duplicate connections are closed', () => {
             wsEndpoint2.connect('ws://127.0.0.1:28501').catch((e) => console.log(e.toString()))
         ])
 
-        // TODO remove when we'll switch to new uWebsocket.js
-        ws1.removeAllListeners()
-        ws2.removeAllListeners()
-
         // TODO enable later
```
harbu, line 41: This todo can be removed I think

test/integration/tracker-assigns-storage-to-empty-stream.test.js
```diff
@@ -96,10 +96,12 @@ describe('tracker assigns storage node to streams on any resend', () => {
     })
 
     afterAll(async () => {
-        await storageNode.stop()
-        await subscriberOne.stop()
-        await subscriberTwo.stop()
-        await tracker.stop()
+        await Promise.all([
```
harbu: Even though awaiting promises sequentially is slightly slower, I find that it better for debugging because you can put breakpoints on individual await lines and see which one exactly is causing the problem.

test/integration/tracker-assigns-storage-node-to-streams.test.js
```diff
@@ -21,24 +22,18 @@ describe('tracker assigns storage node to streams', () => {
...
-
-        await wait(1000)
     })
 
     afterAll(async () => {
-        await storageNode.stop()
-        await subscriberOne.stop()
-        await subscriberTwo.stop()
-        await tracker.stop()
+        await Promise.all([
```
harbu: Even though awaiting promises sequentially is slightly slower, I find that it better for debugging because you can put breakpoints on individual await lines and see which one exactly is causing the problem.

test/integration/l1-resend-requests.test.js
```diff
@@ -94,7 +97,7 @@ describe('resend requests are fulfilled at L1', () => {
         const stream = contactNode.requestResendFrom(
             'streamId',
             0,
-            'requestId',
+            'requestId2',
```
harbu: Probably not needed since new contactNode started before each test?

test/integration/l1-resend-requests.test.js
```diff
@@ -111,7 +114,7 @@ describe('resend requests are fulfilled at L1', () => {
         const stream = contactNode.requestResendRange(
             'streamId',
             0,
-            'requestId',
+            'requestId3',
```
harbu: Same as above

test/integration/l1-resend-requests.test.js
```diff
@@ -72,15 +73,17 @@ describe('resend requests are fulfilled at L1', () => {
...
 
-    afterAll(async () => {
+    afterEach(async () => {
         await contactNode.stop()
         await tracker.stop()
     })
 
     test('requestResendLast', async () => {
-        const stream = contactNode.requestResendLast('streamId', 0, 'requestId', 10)
+        const stream = contactNode.requestResendLast('streamId', 0, 'requestId1', 10)
```
harbu: Prolly can keep original `requestId` since contactNode is now re-instantiated before each test.

test/integration/l3-resend-requests.test.js
```diff
@@ -108,22 +108,14 @@ describe('resend requests are fulfilled at L3', () => {
...
-    afterAll(async () => {
+    afterEach(async () => {
         await tracker.stop()
         await contactNode.stop()
         await neighborOne.stop()
         await neighborTwo.stop()
         await storageNode.stop()
     })
 
-    beforeEach(() => {
```
harbu, line 119: Actually this is needed otherwise we are testing L2. That being said, I found a more deterministic way to do it in WebRTC  branch. Maybe add same thing here?
https://github.com/streamr-dev/network/pull/385/files#diff-38a6c48b5d7654159065a03ee5236c20R109-R114

harbu (changes requested):

src/connection/WsEndpoint.js
mirpo: Both libs have close method, and only `ws` has terminate. So I check this method to understand which socket I currently use.

src/connection/WsEndpoint.js
mirpo: Yes

src/connection/WsEndpoint.js
mirpo: uws is not child of EventEmitter.

src/connection/WsEndpoint.js
mirpo: To make it more async, we already know that it helps. Check [this](https://stackoverflow.com/questions/15349733/setimmediate-vs-nexttick/15349865#15349865)

src/logic/Node.js
mirpo: Tracker should not send empty lists. It means that tracker is malicious or broken

src/logic/Node.js
mirpo: We close all connections in endpoint.

src/logic/Node.js
mirpo: I'll create separate PR which will change the way we add storage into overlay.

harbu: Let N1 be a storage node and subscribed to stream S1. Let's say a new node, N2, joins and subscribes to stream S1. Now the current topology is:
```js
{
   N1: {
       S1: [N2]
   },
   N2: {
      S1: [N1]
   }
}
```

Now, if N2 unsubscribes from S1, tracker will need to send empty list to storage node N1, right?
```
{ stream: S1, nodeAddresses: [] }
```

harbu (approved):

PR #393, opened 2020-02-27
Update travis

mirpo:


juslesan (approved):

.travis.yml
```diff
@@ -1,13 +1,15 @@
...
+- 13.9.0
+git:
+  depth: false
 branches:
   only:
-  - master
-  - "/^v\\d+\\.\\d+(\\.\\d+)?(-\\S*)?$/"
+    - master
+    - "/^v\\d+\\.\\d+(\\.\\d+)?(-\\S*)?$/"
+    - update_uws
```
harbu: Do we still need `update_uws`?

.travis.yml
```diff
@@ -1,13 +1,15 @@
 language: node_js
 node_js:
-- 10.17.0
-sudo: true
+- 13.9.0
+git:
+  depth: false
```
harbu: ?

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Use official node runtime as base image
-FROM node:10.17.0-alpine
+FROM node:13
```
harbu: Btw... I think this Dockerfile can be removed since we are only using broker now?

README.md
```diff
@@ -28,7 +28,7 @@ base.
 
 ## Installation
 
-Prerequisites: [Node.js](https://nodejs.org/) (`^8.10.0`, `^10.13.0`, or `>=11.10.1`), npm version 6+.
+Prerequisites: [Node.js](https://nodejs.org/) (`>=^10.13.0`), npm version 6+.
```
harbu: Should we say exact version because that is what we are actually testing: 13.9.0

Dockerfile
mirpo: Ok

.travis.yml
mirpo: I'll remove

.travis.yml
mirpo: No, it's only for this branch. I'll remove

README.md
mirpo: Ok

PR #396, opened 2020-03-05
Update connection logic and force terminate dead connections

mirpo:


juslesan (approved):

src/connection/WsEndpoint.js
```diff
@@ -152,9 +153,9 @@ class WsEndpoint extends EventEmitter {
                         })
                     } else {
                         this.metrics.inc(`send:failed:readyState=${ws.readyState}`)
-                        this.debug('sent failed because readyState of socket is %d', ws.readyState)
+                        this.debug(`sent to ${recipientAddress} failed because readyState of socket is ${ws.readyState}`)
```
harbu: `sent` -> `send`

src/connection/WsEndpoint.js
```diff
@@ -186,13 +187,13 @@ class WsEndpoint extends EventEmitter {
                             } else {
                                 this.metrics.inc('send:success')
                                 this.debug('sent to %s message "%s"', recipientAddress, message)
-                                resolve()
+                                resolve(this.peerBook.getPeerId(recipientAddress))
```
harbu: `resolve(recipientId)`?

hpihkala:
@mirpo How about adding a test?

src/connection/WsEndpoint.js
```diff
@@ -114,8 +114,9 @@ class WsEndpoint extends EventEmitter {
 
                 if (lastReadyState != null && lastReadyState === ws.readyState) {
                     try {
-                        console.error('terminating connection...')
-                        ws.terminate()
+                        console.error(`closing connection to ${address}...`)
+                        // force close dead of connection
+                        this._onClose(address, this.peerBook.getPeerInfo(address), 1006, 'dead-connection')
```
harbu: But we are now not terminating socket? Method `onClose` only deletes stuff from maps, objects etc. but doesn't terminate connection?

src/logic/Node.js
```diff
@@ -339,27 +352,29 @@ class Node extends EventEmitter {
...
-            return this.protocols.nodeToNode.sendSubscribe(node, streamId).then(() => {
-                this.streams.addInboundNode(streamId, node)
-                this.streams.addOutboundNode(streamId, node)
-
-                // TODO get prove message from node that we successfully subscribed
-                this.emit(events.NODE_SUBSCRIBED, {
-                    streamId,
-                    node
-                })
+        return this.protocols.nodeToNode.sendSubscribe(node, streamId).then((nodeId) => {
```
harbu, line 355: If I understand correctly now subscribe messages may get resent to nodes, leading to some duplicate subscribe messages sent? But this is probably better than being unsure if subscribe has gone through, so LGTM.

src/connection/WsEndpoint.js
mirpo: moved back, but when I was testing, it didn't help, and it doesn't trigger any event...

src/logic/Node.js
mirpo: Yeah, force-subscribe :)
If we'll make subscribe acknowledgement - we can bring back this check later

test/integration/killing-dead-connections.test.js
```diff
@@ -0,0 +1,76 @@
...
+        node1 = await startNetworkNode(LOCALHOST, port1, 'node1')
+        node2 = await startNetworkNode(LOCALHOST, port2, 'node2')
+
+        node1.subscribe(s1, 0)
+        node2.subscribe(s1, 0)
+
+        node1.addBootstrapTracker(tracker.getAddress())
+        node2.addBootstrapTracker(tracker.getAddress())
+
+        await Promise.race([
```
harbu: Are we sure we want `race` here instead of `all`?

test/integration/killing-dead-connections.test.js
```diff
@@ -0,0 +1,76 @@
...
+    afterEach(async () => {
+        allSettled([
+            node1.stop(),
+            node2.stop(),
+            tracker.stop()
+        ])
+    })
+
+    it('if we find dead connection, we force close it', async () => {
+        await Promise.race([
```
harbu: Same as above.

test/integration/killing-dead-connections.test.js
```diff
@@ -0,0 +1,89 @@
...
+    let node1
+    const port1 = 43971
+
+    let node2
+    const port2 = 43972
+
+    const s1 = 'stream-1'
+
+    beforeEach(async () => {
+        tracker = await startTracker(LOCALHOST, trackerPort, 'tracker')
```
harbu: Btw the core feature that is being tested here is in `WsEndpoint` right? I think you could make the test much simpler an avoid having to access members via multi-level dots (`a.b.c.d.property`) by using `startEndpoint` instead of starting tracker and nodes and subscribing and all that. For example, like in this test: https://github.com/streamr-dev/network/blob/master/test/integration/passing-address-between-ws-endpoints.test.js

harbu (approved):

harbu:
LGTM! 🍺

PR #397, opened 2020-03-11
Use only WS codes

mirpo:


src/connection/WsEndpoint.js
```diff
@@ -163,9 +163,7 @@ class WsEndpoint extends EventEmitter {
             } catch (e) {
                 this.metrics.inc('send:failed')
                 console.error('sending to %s failed because of %s, readyState is', recipientAddress, e, ws.readyState)
-                if (ws.readyState === 2 || ws.readyState === 3) {
-                    ws.terminate()
-                }
+                ws.terminate()
```
harbu, line 166: I think we may want to be careful about terminating a "connecting" socket? Maybe leave `if(ws.readyState !== ws.CONNECTING)`?

src/connection/WsEndpoint.js
```diff
@@ -202,9 +200,7 @@ class WsEndpoint extends EventEmitter {
                 } catch (e) {
                     this.metrics.inc('send:failed')
                     console.error('sending to %s failed because of %s, readyState is', recipientAddress, e, ws.readyState)
-                    if (ws.readyState === 2 || ws.readyState === 3) {
-                        ws.terminate()
-                    }
+                    ws.terminate()
```
harbu, line 203: Same as above.

harbu (commented):
LGTM; one question about readyState "connecting" when sending

src/connection/WsEndpoint.js
mirpo: No, we add `ws` to connections only after 'open' or 'connection' event. So ws.readState **must** be => ws.OPEN

harbu: Ah yes, indeed.

harbu (approved):

PR #398, opened 2020-03-11
Remove async in sending instructions, refactor send method

mirpo:


src/connection/WsEndpoint.js
```diff
@@ -135,80 +135,62 @@ class WsEndpoint extends EventEmitter {
 
     sendSync(recipientId, message) {
         const recipientAddress = this.resolveAddress(recipientId)
-        if (!this.isConnected(recipientAddress)) {
+        if (!this.isConnected(recipientAddress) || !this.connections.has(recipientAddress)) {
```
harbu: But isn't `this.isConnected()` already effectively `this.connections.has`?

src/connection/WsEndpoint.js
```diff
@@ -135,80 +135,62 @@ class WsEndpoint extends EventEmitter {
...
+                this._socketSend(ws, message, recipientId, recipientAddress)
+            })
         }
     }
 
     send(recipientId, message) {
         const recipientAddress = this.resolveAddress(recipientId)
         return new Promise((resolve, reject) => {
-            if (!this.isConnected(recipientAddress)) {
+            if (!this.isConnected(recipientAddress) || !this.connections.has(recipientAddress)) {
```
harbu: Same as above

harbu (changes requested):
Nice! See comments, I think you don't need new conditions in the if statements?

src/connection/WsEndpoint.js
mirpo: Good point!

harbu (approved):

PR #401, opened 2020-03-16
Refactor storage logic

mirpo:


src/logic/Tracker.js
```diff
@@ -45,20 +47,23 @@ module.exports = class Tracker extends EventEmitter {
...
         if (isStorage) {
-            this.storageNodes.set(source, status)
+            this.storageNodes.set(source, streams)
         }
 
-        this._updateNode(source, status.streams)
-        this._formAndSendInstructions(source, status.streams)
-        this._formAndSendInstructionsToStorages()
+        this._createNewOverlayTopologies(streams)
+        this._updateAllStorages(streams)
```
harbu: method doesn't have parameters

src/logic/Tracker.js
```diff
@@ -70,28 +75,25 @@ module.exports = class Tracker extends EventEmitter {
...
         if (!foundStorageNodes.length) {
-            const randomStorage = this._getRandomStorage()
-
-            if (randomStorage) {
-                foundStorageNodes.push(randomStorage)
-            }
+            foundStorageNodes = foundStorageNodes.concat([...this.storageNodes.keys()])
         }
 
+        this._updateAllStorages(requestStreams)
```
harbu: method doesn't have parameters

src/logic/Tracker.js
```diff
@@ -104,58 +106,63 @@ module.exports = class Tracker extends EventEmitter {
...
+            this.storageNodes.set(storageId, updateStreams)
+            this._updateNode(storageId, updateStreams)
+        })
+    }
 
-        // Add or update
-        Object.entries(streams).forEach(([streamKey, { inboundNodes, outboundNodes }]) => {
-            if (this.overlayPerStream[streamKey] == null) {
-                this.overlayPerStream[streamKey] = this._createNewOverlayTopology()
+    _createNewOverlayTopologies(streams) {
```
harbu, line 132: Maybe change parameter to be array of streamIds? Then you can just
```
streamIds.forEach((streamId) => ...
```

Related: see things marked with (*)

src/logic/Tracker.js
```diff
@@ -70,28 +75,25 @@ module.exports = class Tracker extends EventEmitter {
         // especially those that aren't actively being subscribed or produced to. Thus on encountering a
         // unknown streamId, we need to create a new topology and assign storage node(s) to it to ensure
         // that resend requests for inactive streams get properly handled.
-        if (this.storageNodes.size && this.overlayPerStream[streamId] == null) {
-            this.overlayPerStream[streamId] = this._createNewOverlayTopology()
-            this._formAndSendInstructionsToStorages()
+        const requestStreams = {}
```
harbu, line 78: (*) Can be simplified, see my comment below

src/logic/Tracker.js
```diff
@@ -70,28 +75,25 @@ module.exports = class Tracker extends EventEmitter {
...
         })
 
-        // TODO: this works for single storage node scenario. How to deal with multiple?
         if (!foundStorageNodes.length) {
-            const randomStorage = this._getRandomStorage()
-
-            if (randomStorage) {
-                foundStorageNodes.push(randomStorage)
-            }
+            foundStorageNodes = foundStorageNodes.concat([...this.storageNodes.keys()])
```
harbu: What about just
```js
foundStorageNodes = [...this.storageNodes.keys()]
```
since foundStorageNodes will be empty in this case?

src/logic/Tracker.js
```diff
@@ -104,58 +106,63 @@ module.exports = class Tracker extends EventEmitter {
...
         })
+    }
 
-        // Remove
-        const currentStreamKeys = new Set(Object.keys(streams))
-        Object.entries(this.overlayPerStream)
-            .filter(([streamKey, _]) => !currentStreamKeys.has(streamKey))
-            .forEach(([streamKey, overlayTopology]) => this._leaveAndCheckEmptyOverlay(streamKey, overlayTopology, node))
+    _updateNode(node, streams) {
+        if (!isEmpty(streams)) {
```
harbu: Hmmm.. this is very subjective, but I find
```
if (isEmpty((streams)) {
    this._removeNode(node)
    return
}
...
...
...
...
...
...
...
...
``` 

easier to read than

```
if (!isEmpty((streams)) {
    ...
    ...
    ...
    ...
    ...
    ...
    ...
    ...
    return
}
this._removeNode(node)
```

test/integration/tracker-assigns-storage-to-empty-stream.test.js
```diff
@@ -1,10 +1,9 @@
 const intoStream = require('into-stream')
 const { StreamMessage } = require('streamr-client-protocol').MessageLayer
-const { waitForEvent, waitForCondition, waitForStreamToEnd } = require('streamr-test-utils')
+const { wait, waitForEvent, waitForCondition, waitForStreamToEnd } = require('streamr-test-utils')
```
harbu: No need for this change since you aren't using `wait`.

src/logic/Tracker.js
mirpo: We get from status message `Object` of streams. So let's keep it.

src/logic/Tracker.js
mirpo: Good point!

src/logic/Tracker.js
harbu: You could in processNodeStatus do

```js
this._createNewOverlayTopologies(Object.keys(streams))
```

harbu (approved):

PR #405, opened 2020-03-20
Fix subscribe logic, update _checkConnections

mirpo:


juslesan (approved):

src/connection/WsEndpoint.js
```diff
@@ -22,10 +22,14 @@ const ab2str = (buf) => Buffer.from(buf).toString('utf8')
...
-    } else {
-        ws.end(code, reason)
+    try {
+        if (ws.terminate !== undefined) {
+            ws.close(code, reason)
+        } else {
+            ws.end(code, reason)
+        }
+    } catch (e) {
+        //
```
harbu: What's this about? Should error be logged?

src/connection/WsEndpoint.js
```diff
@@ -39,11 +43,15 @@ const getBufferedAmount = (ws) => {
...
-        ws.close()
+    try {
+        // only ws/ws lib has terminate method
+        if (ws.terminate !== undefined) {
+            ws.terminate()
+        } else {
+            ws.close()
+        }
+    } catch (e) {
+        //
```
harbu: What's this about? Should error be logged?

src/connection/WsEndpoint.js
```diff
@@ -138,38 +139,22 @@ class WsEndpoint extends EventEmitter {
...
-                    } finally {
-                        this.lastCheckedReadyState.delete(address)
-                    }
-                }
-            } else {
-                this.lastCheckedReadyState.delete(address)
+            try {
+                ws.ping('ping')
+            } catch (e) {
+                terminateWs(ws)
```
harbu, line 154: Should we log error?

src/connection/WsEndpoint.js
```diff
@@ -417,6 +421,11 @@ class WsEndpoint extends EventEmitter {
             // eslint-disable-next-line no-param-reassign
             ws.address = address
 
+            if (this.isConnected(address)) {
```
harbu, line 421: I wonder how this interacts with the duplicate detection logic we have in method `_onNewConnection`.

src/logic/Node.js
```diff
@@ -181,6 +181,15 @@ class Node extends EventEmitter {
             await allSettled(connectedNodes.map((nodeId) => this._subscribeToStreamOnNode(nodeId, streamId))).then((results) => {
                 results.forEach((result) => {
                     if (result.status === 'fulfilled') {
+                        const node = result.value
```
harbu: What's the reasoning behind moving this logic here? To me it really seems that adding the inbound/outbound nodes and emitting the event belongs in the `_subscribeToStreamOnNode` method.

test/integration/network-stabilization.test.js
```diff
@@ -0,0 +1,45 @@
...
+const allSettled = require('promise.allsettled')
+
+const { startNetworkNode, startTracker } = require('../../src/composition')
+const { LOCALHOST } = require('../util')
+
+describe('check network stabilization', () => {
+    let tracker
+    const trackerPort = 39000
+
+    const nodes = []
```
harbu: It's better to initialise in `beforeEach` to avoid surprises in future if someone decides to add a new test case to this suite.

test/integration/network-stabilization.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+    afterEach(async () => {
+        for (let i = 0; i < MAX_NODES; i++) {
+            // eslint-disable-next-line no-await-in-loop
+            await nodes[i].stop()
+        }
+        await tracker.stop()
+    }, 10000)
+
+    it('expect to _formAndSendInstructions not to be called when topology is stable', async () => {
```
harbu: "expect <del>to</del> _formAndSendInstructions not to be called..."

test/integration/network-stabilization.test.js
```diff
@@ -0,0 +1,45 @@
...
+        await wait(10000)
```
harbu: I'd suggest inserting a check that the function `_formAndSendInstructions` exists before the await, e.g.,
```js
expect(this._formAndSendInstructions).toBeInstanceOf(Function)
```

This is in case we decide to re-name the function in the future or whatever.

harbu (changes requested):
Have some questions on some of the changes and a few minor improvement ideas.

src/connection/WsEndpoint.js
mirpo: We don't need to log this error, anyway we're closing socket

src/connection/WsEndpoint.js
mirpo: We don't need to log this error, anyway we're closing socket

mirpo: `//` for eslint

src/connection/WsEndpoint.js
mirpo: We had before this logic in `verifyClient`. But `uws` doesn't have this method.

src/logic/Node.js
mirpo: I just removed duplicate code.
Because 
```
async _subscribeToStreamOnNode(node, streamId) {
        if (this.streams.hasInboundNode(streamId, node) && this.streams.hasOutboundNode(streamId, node)) {
            return node
        }

        return this.protocols.nodeToNode.sendSubscribe(node, streamId).then((nodeId) => {
            return nodeId
        })
    }
```
here we need to trigger NODE_SUBSCRIBED in both cases to propagate buffered messages

test/integration/network-stabilization.test.js
mirpo: ok

test/integration/network-stabilization.test.js
mirpo: good point

src/connection/WsEndpoint.js
harbu: I think the problem is that the catch is too wide and it could hide other errors than what was intended. For example, what if `WsEndpoint` had some bug that causes this function to be called as follows `closeWs(undefined, 101, 'lorem')` Then `ws.end()`  would throw an error (because `ws = undefined`) but this would then get hidden from us.

Maybe the intention is to do something like the following (image this was Java or C++ for a minute):
```java
try {
  ws.end()
} catch (ReadyStateException e) {
   // nop
} catch (Error e) {
    console.error(e)
}
```

I think you can do something similar in Node.js
```
try {
} catch (e) {
   if (e instance of ReadyStateError) {
   } else {
      throw e
   }
}
```

Nowadays you can also use https://nodejs.org/api/errors.html#errors_error_code I guess?

src/connection/WsEndpoint.js
harbu: Same comment as above

src/connection/WsEndpoint.js
harbu: ?

src/logic/Node.js
harbu: I don't understand how this removes duplicate code. I just see the code moved from one place to another?

The way the code behaves is slightly different now, is this intentional? Whereas before it would add inbound/outboud nodes and emit event as soon as the subscribe message went through, now it will wait for subscribe messages for all nodes to go through (or to fail) via `allSettled` and only then proceed to add inbound/outbound nodes and emit events.

src/connection/WsEndpoint.js
harbu: I ran `npm run network 100 2> test-**.txt` for three variants 5 times:
1) one without this change
2) one with this change
3) one with this change but I remove other check from `_onNewConnection`

Firstly, we get actually race conditions _without_ this change (1) (pairs of nodes continually connecting & disconnecting without stop).

Both options (2) and (3) work much better, actually option (3) leads to lower amount of debug messages? So I'm wondering would it actually make sense to keep this and _remove_ the duplicate detection logic inside `_onNewConnection`?

src/logic/Node.js
harbu: Assuming this is intentional, I think we should just then remove method `_subscribeToStreamOnNode`  and move stuff here? The problem I see is that method `_subscribeToStreamOnNode` no longer does not do what it claims. If I need to call `_subscribeToStreamOnNode` in future from somewhere else in this class, I will need to remember to do addInboundNodes/addOutBoundNodes/emit otherwise it won't properly subscribe.

harbu (changes requested):

src/connection/WsEndpoint.js
mirpo: Agree, in separate PR

mirpo:
About `_subscribeToStreamOnNode`. I check inside function ` if (this.streams.hasInboundNode(streamId, node) && this.streams.hasOutboundNode(streamId, node))` 
this `if` means that 
```
this.streams.addInboundNode(streamId, node)
this.streams.addOutboundNode(streamId, node)
```
were called before. If you want for readability, I can add calls in `first if`, but they'll be useless


```
async _subscribeToStreamOnNode(node, streamId) {
        if (this.streams.hasInboundNode(streamId, node) && this.streams.hasOutboundNode(streamId, node)) {
            this.emit(events.NODE_SUBSCRIBED, {
                streamId,
                node
            })
            return node
        }

        return this.protocols.nodeToNode.sendSubscribe(node, streamId).then((nodeId) => {
            this.streams.addInboundNode(streamId, node)
            this.streams.addOutboundNode(streamId, node)

            this.emit(events.NODE_SUBSCRIBED, {
                streamId,
                node
            })
            return nodeId
        })
    }
```

src/connection/WsEndpoint.js
mirpo: ws/ws emits 
```
const err = new Error(message);
  Error.captureStackTrace(err, abortHandshake);
```
uws emits standard error too

src/connection/WsEndpoint.js
harbu: Yup

src/connection/WsEndpoint.js
harbu: Meh, sort of annoying decision from library author(s). Oh well. I see you changed to `console.error` so resolving.

harbu:
What about

```js
async _subscribeToStreamOnNode(node, streamId) {
    if (!this.streams.hasInboundNode(streamId, node) || !this.streams.hasOutboundNode(streamId, node)) {
        await this.protocols.nodeToNode.sendSubscribe(node, streamId)
    }

    this.streams.addInboundNode(streamId, node)
    this.streams.addOutboundNode(streamId, node)

    this.emit(events.NODE_SUBSCRIBED, {
        streamId,
        node
     })

    return node
}
```

Or if you'd rather avoid `await` then maybe

```js
_subscribeToStreamOnNode(node, streamId) {
    let p
    if (this.streams.hasInboundNode(streamId, node) && this.streams.hasOutboundNode(streamId, node)) {
        p = Promise.resolve(node)
    } else {
        p = this.protocols.nodeToNode.sendSubscribe(node, streamId)
    }

    return p.then(() => {
        this.streams.addInboundNode(streamId, node)
        this.streams.addOutboundNode(streamId, node)

        this.emit(events.NODE_SUBSCRIBED, {
            streamId,
            node
        })

        return node
    })
}
```

Would these work?

harbu (changes requested):

harbu (approved):
LGTM

PR #409, opened 2020-04-17
Multi tracker

mirpo:


src/logic/Node.js
```diff
@@ -310,27 +323,35 @@ class Node extends EventEmitter {
...
-            this.trackers.forEach((tracker) => this._sendStatus(tracker))
+    _sendStreamStatus(streamId) {
+        const streamKey = streamId.key()
+        clearTimeout(this.sendStatusTimeout.get(streamKey))
+
+        const timeout = setTimeout(() => {
+            const trackerId = this.trackersRing.get(streamKey)
+            this._sendStatus(trackerId)
         }, this.opts.sendStatusToAllTrackersInterval)
+        this.sendStatusTimeout.set(streamKey, timeout)
```
harbu: Have you thought about using `trackerId` as key for `this.sendStatusTimeout` instead? I'm guessing now if two streams are assigned to same tracker, we will have two timeouts for same tracker. Could result in less status messages being sent if we would use `trackerId` as key instead?

src/logic/Node.js
```diff
@@ -310,27 +323,35 @@ class Node extends EventEmitter {
...
+        const timeouts = [...this.sendStatusTimeout.values()]
+        timeouts.forEach((timeout) => clearTimeout(timeout))
+
         this._clearConnectToBootstrapTrackersInterval()
         this.messageBuffer.clear()
         return this.protocols.nodeToNode.stop()
     }
 
-    _getStatus() {
+    _getStatus(tracker, trackersRing) {
```
harbu: You already have `trackersRing` in `this`. No need for parameter `trackersRing` in method.

test/integration/network-stabilization.test.js
```diff
@@ -37,11 +34,17 @@ describe('check network stabilization', () => {
...
     }, 40000)
 
-    it('expect _formAndSendInstructions not to be called when topology is stable', async () => {
-        await wait(15000)
-        const spy = jest.spyOn(tracker, '_formAndSendInstructions').mockImplementation(() => {})
-        await wait(10000)
-        expect(spy).not.toHaveBeenCalled()
-        jest.restoreAllMocks()
-    }, 40000)
+    it('netowork must become stable in less than 15 seconds', async (done) => {
```
harbu: netowork

harbu (approved):
LGTM

PR #412, opened 2020-04-27
Ping pong with RTT

mirpo:


src/connection/WsEndpoint.js
```diff
@@ -135,22 +135,41 @@ class WsEndpoint extends EventEmitter {
...
-                ws.ping('ping')
+                // didn't get "pong" in pingInterval
+                if (ws.respondedPong !== undefined && !ws.respondedPong) {
+                    throw Error('ws is not active')
+                }
+
+                // eslint-disable-next-line no-param-reassign
+                ws.respondedPong = false
+                ws.rttStart = Date.now()
+                ws.rtt = ws.rttStart - ws.rttEnd
```
harbu: Shouldn't this be moved to `pong`?

```js
pong: (ws) => {
                const connection = this.connections.get(ws.address)

                if (connection) {
                    this.debug(`received from ${ws.address} "pong" frame`)
                    connection.respondedPong = true
                    connection.rtt = Date.now() - connection.rttStart
                }
```

src/connection/WsEndpoint.js
```diff
@@ -477,6 +497,14 @@ class WsEndpoint extends EventEmitter {
             setImmediate(() => this.onReceive(peerInfo, address, message))
         })
 
+        ws.on('ping', () => {
```
harbu: Don't think this is needed though? Why should a received ping affect this node's RTT calculation? I think the RTT calculation from node A to node B should be independent from calculation from node B to node A.

src/logic/Node.js
```diff
@@ -342,13 +342,16 @@ class Node extends EventEmitter {
     _sendStreamStatus(streamId) {
         const streamKey = streamId.key()
         const trackerId = this.trackersRing.get(streamKey)
-        clearTimeout(this.sendStatusTimeout.get(trackerId))
 
-        const timeout = setTimeout(() => {
-            this._sendStatus(trackerId)
-        }, this.opts.sendStatusToAllTrackersInterval)
+        if (trackerId) {
```
harbu, line 346: Don't understand? When would a trackerId not be returned for a given streamKey? It always has to be assigned to some tracker?

src/connection/WsEndpoint.js
harbu: Otherwise we only call setInterval() every `pingInterval` so this will lead to error in measurement?

harbu (changes requested):

src/logic/Node.js
mirpo: for example:
```js
startNetworkNode(program.ip, program.port, id).then((subscriber) => {
    console.log('started subscriber id: %s, port: %d, ip: %s, trackers: %s, streamId: %s, metrics: %s',
        id, program.port, program.ip, program.trackers.join(', '), program.streamId, program.metrics)

    subscriber.subscribe(program.streamId, 0)
    program.trackers.map((trackerAddress) => subscriber.addBootstrapTracker(trackerAddress))
```

we subscribe, but not connected yet

src/connection/WsEndpoint.js
mirpo: Good point

src/connection/WsEndpoint.js
mirpo: We have two kinds of connections -> inbound and outbound, that's we sometimes we ping `uws` socket, sometimes `ws`. So uws `pong` will be arrive to `ws server`, `ws` will be triggered in this event.

src/logic/Node.js
harbu: Alright so basically when there are no trackers added.

src/connection/WsEndpoint.js
harbu: But is this event triggered when we _receive_ a ping? The only thing I think that should happen when receiving ping is to respond pong.

1. Node A: record `startTime`, send `ping` to Node B
2. Node B: receive `ping` from Node A, send `pong` to Node A
3. Node A: receive `pong` from Node B, calculate `rtt = startTime - now()`

juslesan (approved):

harbu (approved):
LGTM!

PR #416, opened 2020-05-14
Add disconnection timer to node

harbu:
Cherry-picked from WebRTC branch.

Upon noticing that there are no shared streams, a node will now wait a pre-determined time (`disconnectionWaitTime`) before actually disconnecting from the other node. If during this time a new connection is formed between the two nodes, or if after the timeout the two nodes have shared streams again, the disconnection will be canceled. This allows nodes to be more resilient in the face of connection churn caused by tracker instructions during network stabilisation.

### DIFF
--- a/src/logic/Node.js
+++ b/src/logic/Node.js
@@ -334,6 +334,8 @@ class Node extends EventEmitter {
         const timeouts = [...this.sendStatusTimeout.values()]
         timeouts.forEach((timeout) => clearTimeout(timeout))
 
+        Object.values(this.disconnectionTimers).forEach((timeout) => clearTimeout(timeout))
+
         this._clearConnectToBootstrapTrackersInterval()
         this.messageBuffer.clear()
         return this.protocols.nodeToNode.stop()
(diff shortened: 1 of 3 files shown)


src/logic/Node.js
```diff
@@ -327,6 +334,8 @@ class Node extends EventEmitter {
         const timeouts = [...this.sendStatusTimeout.values()]
         timeouts.forEach((timeout) => clearTimeout(timeout))
 
+        Object.values(this.disconnectionTimers).forEach((timeout) => clearTimeout(timeout))
```
mirpo, line 337: Why do you clean all disconnectionTimers? You already cleared needed above `this._clearDisconnectionTimer(node)`? I guess all other timers must trigger?

mirpo (changes requested):

src/logic/Node.js
harbu: This is `stop` method so we want to make sure all timers are cleared so that no hanging timeouts are alive. Not sure I understand what you mean with 2nd point?

mirpo (approved):

PR #417, opened 2020-06-02
Upgrade protocol library to 5.x

hpihkala:
Start using the new major version of the protocol library, `5.x`. Main changes in the new library version:

- The constructor interfaces of ControlLayer messages and `StreamMessage`s have been unified (the changes in this PR are mostly related to this refactoring)
- Adds support for ControlLayer version `2` and drops support for version `0`
- The serialization/deserialization of messages is a lot faster
- The lib contains utility classes for message validation (not used in `network` yet, but will be used in the future)

### DIFF
--- a/src/protocol/NodeToNode.js
+++ b/src/protocol/NodeToNode.js
@@ -1,5 +1,6 @@
 const { EventEmitter } = require('events')
 
+const { v4: uuidv4 } = require('uuid')
 const { ControlLayer } = require('streamr-client-protocol')
 
 const encoder = require('../helpers/MessageEncoder')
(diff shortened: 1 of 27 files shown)


mirpo (approved):
LGFM, much easier to read code.

src/protocol/NodeToNode.js
```diff
@@ -43,15 +44,26 @@ class NodeToNode extends EventEmitter {
     }
 
     sendData(receiverNodeId, streamMessage) {
-        this.endpoint.sendSync(receiverNodeId, encoder.wrapperMessage(ControlLayer.BroadcastMessage.create(streamMessage)))
+        this.endpoint.sendSync(receiverNodeId, encoder.wrapperMessage(new ControlLayer.BroadcastMessage({
+            requestId: '', // TODO: how to echo here the requestId of the original SubscribeRequest?
```
harbu, line 48: Hmm.. good question. I think the empty value suffices for now.

For a solution with non-empty requestId(s), I think we would want to store the requestId(s) in, e.g., the `StreamManager` class in logic layer. Thus logic layer would pass down the `requestId` as an method argument.

harbu (approved):
LGTM; glad we went with the object parameters pattern as opposed to positional parameters as it makes the library more useable imo.

src/protocol/NodeToNode.js
hpihkala: Logged this technical debt as a [ticket](https://streamr.atlassian.net/browse/NET-120). Having the correct `requestId` there has little practical significance at the moment, so assigning only minor priority to it.

PR #418, opened 2020-06-05
scripts for studying overlay toplogy randomness and performance

harbu:
Two "tests" for `OverlayTopology`:
- Performance test test/benchmarks/overlay-topology-performance.test.js
- Script for studying randomness test/benchmarks/overlay-topology-randomness.js


mirpo (approved):
💻 🎖️

PR #419, opened 2020-06-17
Update libs and uWebSockets

mirpo:
uWebSockets added upgrade listener in server, so I decided to remove my hack in code. 
But they have bug and it's not possible to use built-in function `writeHeader`. But it's possible to write headers using `writeStatus`


src/connection/WsEndpoint.js
```diff
@@ -99,21 +99,27 @@ class WsEndpoint extends EventEmitter {
...
+            upgrade: (res, req, context) => {
+                res.writeStatus(`101 Switching Protocols\nstreamr-peer-id: ${this.peerInfo.peerId}\nstreamr-peer-type: ${this.peerInfo.peerType}`)
+
+                /* This immediately calls open handler, you must not use res after this call */
+                res.upgrade({
+                    query: req.getQuery(),
+                    peerId: req.getHeader('streamr-peer-id'),
+                    peerType: req.getHeader('streamr-peer-type'),
+                },
+                /* Spell these correctly */
```
harbu, line 115: What is this stuff below?

harbu (approved):

src/connection/WsEndpoint.js
mirpo: If we add `upgrade` handler, we need to call explicitly `res.upgrade`.

juslesan (approved):

PR #421, opened 2020-06-18
Add try/catch to encoder

mirpo:
Now `encoder` doesn't crash when:
- gets not `JSON`
- unknown message type

### DIFF
--- a/src/helpers/MessageEncoder.js
+++ b/src/helpers/MessageEncoder.js
@@ -21,7 +21,14 @@ const encode = (type, payload) => {
 }
 
 const decode = (source, message) => {
-    const { code, payload } = JSON.parse(message)
+    let code
+    let payload
+
+    try {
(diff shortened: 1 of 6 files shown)


src/helpers/MessageEncoder.js
```diff
@@ -21,7 +21,14 @@ const encode = (type, payload) => {
...
 
 const decode = (source, message) => {
-    const { code, payload } = JSON.parse(message)
+    let code
+    let payload
+
+    try {
+        ({ code, payload } = JSON.parse(message))
+    } catch (e) {
+        return undefined
```
harbu, line 30: I think you need to add extra logic to code that call this function to check for `undefined` otherwise they will crash.

juslesan (approved):

harbu (approved):

PR #424, opened 2020-07-01
Instruction counter

harbu:
Add counters to instruction messages and status messages.

- Tracker keeps a counter per `(nodeId, streamId)` pair. Tracker increments this counter each time it sends an instruction to `nodeId` for stream `streamId`.
- Upon receiving an instruction from the tracker, node `nodeId` will execute the instructions and then update its internal counter for `streamId` to match the one provided in the instruction. 
- When it is time for node `nodeId` to send a status message to tracker, for each stream, it will include the associated counter value it has stored internally.
- Upon receiving a status message from node `nodeId`, tracker will filter the streams in the message according to counters. The tracker will only handle those streams whose associated counter in the status message is (greater than or) equal to the tracker's current internal counter.

### DIFF
--- a/src/helpers/MessageEncoder.js
+++ b/src/helpers/MessageEncoder.js
@@ -38,6 +38,7 @@ const decode = (source, message) => {
             return new InstructionMessage(
                 new StreamIdAndPartition(payload.streamId, payload.streamPartition),
                 payload.nodeAddresses,
+                payload.counter,
                 source
             )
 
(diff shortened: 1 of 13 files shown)


harbu:
Re-based to master (instead of webrtc)

mirpo (approved):
🎖️

juslesan (approved):

PR #425, opened 2020-07-03
Rtt status

juslesan:
- Include RTTs of connections in status messages

- Store connection RTTs of the overlay in tracker


src/logic/Node.js
```diff
@@ -344,7 +344,8 @@ class Node extends EventEmitter {
     _getStatus(tracker) {
         return {
             streams: this.streams.getStreamsWithConnections(tracker, this.trackersRing),
-            started: this.started
+            started: this.started,
+            rtts: this.protocols.nodeToNode.endpoint.getRtts()
```
harbu: Think you can shorten this to `this.protocols.nodeToNode.getRtts()` since you added a method there.

test/integration/tracker-node-status.test.js
```diff
@@ -66,4 +67,30 @@ describe('check status message flow between tracker and two nodes', () => {
...
+                    expect(statusMessage.getStatus().rtts[nodeOne.opts.id]).toBeGreaterThanOrEqual(0)
+                }
+                receivedTotal += 1
+                if (receivedTotal === 2) {
+                    done()
+                }
+            })
+            nodeOne.subscribe(streamId2, 0)
+            nodeTwo.subscribe(streamId2, 0)
+        }, 2000)
```
harbu: Btw this test passed for  me without the `setTimeout()`. So in furtherance of faster test suite execution can we leave the `setTimeout` out or is there some need for it?

juslesan: It seems that the timeout is unnecessary. I had it there to guarantee that the first WS ping pong has passed so that the rtt value is defined. It seems that it is not necessary when there are only 2 nodes.

harbu (approved):

mirpo (approved):

PR #428, opened 2020-07-08
Instruction Throttling

harbu:
In this PR the logic of instruction handling on node side is changed as follows.

1. Node maintains an instance of `InstructionThrottler` during its lifecycle.
2. Upon receiving an instruction from tracker, it calls `InstructionThrottler#add` to enqueue the instruction.
3. `InstructionThrottler` invokes a callback supplied by Node (i.e. `Node#handleTrackerInstruction`) whenever a new instruction should be handled.

The class `InstructionThrottler` makes sure that no more than one instruction is handled at a time. While an instruction is being handled and a new instruction arrives, it enqueues the new instruction. If several instructions with the same streamId arrive during handling, it will keep only the latest one around (basically the one with the highest counter).

## Some considerations
- This does add complexity to the instruction handling logic. It would be nice if we could somehow check if this change brings more stability to the network.
-  A previous version of this change had a less strict invariant. It allowed for more than one instruction to be handled at a time given that they operate on different streams. (So basically max 1 instruction under handling at a time per streamId.)

### DIFF
--- a/src/logic/InstructionThrottler.js
+++ b/src/logic/InstructionThrottler.js
@@ -9,7 +9,6 @@ module.exports = class InstructionThrottler {
         this.handling = false
         this.handleFn = handleFn
         this.queue = {} // streamId => instructionMessage
-        this.timeout = null
     }
 
     add(instructionMessage) {
@@ -25,11 +24,6 @@ module.exports = class InstructionThrottler {
 
     isIdle() {
         return !this.handling
-    }
-
-    clear() {
-        this.queue = {}
-        clearTimeout(this.timeout)
     }
 
     async _invokeHandleFnWithLock() {
--- a/src/logic/InstructionThrottler.js
+++ b/src/logic/InstructionThrottler.js
@@ -1,0 +1,56 @@
+/**
+ * InstructionThrottler makes sure that
+ *  1. no more than one instruction is handled at a time
+ *  2. any new instructions arriving while an instruction is being handled are queued in a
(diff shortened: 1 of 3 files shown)


src/logic/InstructionThrottler.js
```diff
@@ -0,0 +1,56 @@
...
+ *  1. no more than one instruction is handled at a time
+ *  2. any new instructions arriving while an instruction is being handled are queued in a
+ *     way where only the most latest instruction per streamId is kept in queue.
+ */
+module.exports = class InstructionThrottler {
+    constructor(handleFn) {
+        this.handling = false
+        this.handleFn = handleFn
+        this.queue = {} // streamId => instructionMessage
+        this.timeout = null
```
mirpo: not in use?

src/logic/InstructionThrottler.js
```diff
@@ -0,0 +1,56 @@
...
+        delete this.queue[streamId]
+    }
+
+    isIdle() {
+        return !this.handling
+    }
+
+    clear() {
+        this.queue = {}
+        clearTimeout(this.timeout)
```
mirpo: not in use?

juslesan (approved):
This change seems to remove the issue of tracker sending instructions with more than 5 connections. Tested with 16, 32, 64, 128, 256 and 512 node stream topologies on emulator.

harbu:
This should be okay, but probably worth fixing memory leak issue before merging this.

mirpo (changes requested):

PR #429, opened 2020-07-09
Status interval

juslesan:


mirpo:
Conflicting files

src/logic/Node.js
```diff
@@ -108,8 +110,10 @@ class Node extends EventEmitter {
 
         this.trackers.add(tracker)
         this.trackersRing.add(tracker)
-
         this._sendStatus(tracker)
+        this.statusIntervals[tracker] = setInterval(() => {
```
timoxley, line 114: Even if it's "not possible" should always clear interval before setting new one. Bugs related to timers like this can be hard to notice and can create highly confusing behaviour, while also being easy to prevent by always clearing the timer before setting it.

Also see this comment about `setInterval`: https://github.com/streamr-dev/network/pull/456/files#r495946918

harbu:
Don't think we have a pressing need for this? So closing for now.

PR #435, opened 2020-08-04
Allow PeerInfo names

juslesan:
Supporting PR for https://github.com/streamr-dev/broker/pull/164

### DIFF
--- a/src/connection/PeerInfo.js
+++ b/src/connection/PeerInfo.js
@@ -5,7 +5,7 @@ const peerTypes = Object.freeze({
 })
 
 class PeerInfo {
-    static newTracker(peerId, peerName = 'TrackerNode') {
+    static newTracker(peerId, peerName) {
         return new PeerInfo(peerId, peerTypes.TRACKER, peerName)
     }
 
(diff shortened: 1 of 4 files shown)


src/connection/PeerInfo.js
```diff
@@ -5,19 +5,19 @@ const peerTypes = Object.freeze({
...
     }
 
-    static newStorage(peerId) {
-        return new PeerInfo(peerId, peerTypes.STORAGE)
+    static newStorage(peerId, peerName) {
+        return new PeerInfo(peerId, peerTypes.STORAGE, peerName)
     }
 
-    constructor(peerId, peerType) {
+    constructor(peerId, peerType, peerName = 'StreamrNetworkNode') {
```
harbu: Let's just use empty string as default name? No name given => use ID as "name".

harbu (commented):
What about exchanging name info with other nodes on connection?

juslesan:
> What about exchanging name info with other nodes on connection?

I will add a separate PR for this

mirpo (approved):

harbu (approved):
LGTM

PR #438, opened 2020-08-06
Peername passing

juslesan:
Passing peerInfo peerNames to PeerBook.
Passing of peerNames of connections to the logic layer should require changes to the protocol

### DIFF
--- a/src/connection/PeerBook.js
+++ b/src/connection/PeerBook.js
@@ -19,18 +19,20 @@ class PeerBook {
         this.idToAddress = {}
         this.addressToId = {}
         this.addressToType = {}
+        this.addressToName = {}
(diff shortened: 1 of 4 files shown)


harbu (approved):

src/connection/PeerBook.js
```diff
@@ -19,18 +19,20 @@ class PeerBook {
         this.idToAddress = {}
         this.addressToId = {}
         this.addressToType = {}
+        this.addressToName = {}
```
mirpo, line 22: Not about this PR, but I think it's time to store all information about node in one object, what do you think @harbu ?

mirpo (approved):

PR #439, opened 2020-08-07
Fix storage for all resends

mirpo:
**IMPORTANT** revert this PR after data migration is done.


harbu (approved):
LGTM; as you said let's try to remember to revert (also tests) when all this is done.

PR #443, opened 2020-08-10
Geolocation

juslesan:
- Location data saved in peerInfo

- Nodes now report defined location data on status to tracker

- Leaves the option for real time updates in the future

- Tracker saves location data and has an endpoint for fetching the data

https://github.com/streamr-dev/broker/pull/172

### DIFF
--- a/bin/tracker.js
+++ b/bin/tracker.js
@@ -97,7 +97,21 @@ function startServer(tracker, endpointServerPort) {
         reply.send(tracker.getTopology(request.params.streamId, request.params.partition))
     })
 
-    //
+    fastify.get('/location/', async (request, reply) => {
+        reply.send(tracker.getAllNodeLocations())
+    })
+
+    fastify.get('/location/:nodeId', async (request, reply) => {
+        if (request.params.nodeId === '') {
+            throw Error('nodeId must not be an empty string')
+        }
+        const location = tracker.getNodeLocation(request.params.nodeId)
+        if (!location) {
+            throw Error('Node not found')
+        }
+        reply.send(location)
+    })
+
     fastify.listen(endpointServerPort, '0.0.0.0', (err, address) => {
         if (err) {
             throw err
--- a/src/composition.js
+++ b/src/composition.js
@@ -9,7 +9,14 @@ const Tracker = require('./logic/Tracker')
 const NetworkNode = require('./NetworkNode')
 const { startEndpoint } = require('./connection/WsEndpoint')
 
-function startTracker(host, port, id = uuidv4(), maxNeighborsPerNode = 4, advertisedWsUrl = null, name) {
+const defaultLocation = {
+    latitude: null,
+    longitude: null,
+    country: null,
+    city: null
+}
+
+function startTracker(host, port, id = uuidv4(), maxNeighborsPerNode = 4, advertisedWsUrl = null, name, location = defaultLocation) {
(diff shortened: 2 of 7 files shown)


bin/tracker.js
```diff
@@ -97,7 +97,21 @@ function startServer(tracker, endpointServerPort) {
...
+        reply.send(tracker.getAllNodeLocations())
+    })
+
+    fastify.get('/location/:nodeId', async (request, reply) => {
+        if (request.params.nodeId === '') {
+            throw Error('nodeId must not be an empty string')
+        }
+        const location = tracker.getNodeLocation(request.params.nodeId)
+        if (!location) {
+            throw Error('Node not found')
```
mirpo, line 110: What it does? Returns `JSON` error? Or some `HTTP` error?

src/composition.js
```diff
@@ -9,7 +9,14 @@ const Tracker = require('./logic/Tracker')
...
-function startTracker(host, port, id = uuidv4(), maxNeighborsPerNode = 4, advertisedWsUrl = null, name) {
+const defaultLocation = {
+    latitude: null,
+    longitude: null,
+    country: null,
+    city: null
+}
+
+function startTracker(host, port, id = uuidv4(), maxNeighborsPerNode = 4, advertisedWsUrl = null, name, location = defaultLocation) {
     const peerInfo = PeerInfo.newTracker(id, name)
```
mirpo: `defaultLocation` is not in use?

src/composition.js
```diff
@@ -23,7 +30,7 @@ function startTracker(host, port, id = uuidv4(), maxNeighborsPerNode = 4, advert
     })
 }
 
-function startNetworkNode(host, port, id = uuidv4(), storages = [], advertisedWsUrl = null, name) {
+function startNetworkNode(host, port, id = uuidv4(), storages = [], advertisedWsUrl = null, name, location = defaultLocation) {
     const peerInfo = PeerInfo.newNode(id, name)
     return startEndpoint(host, port, peerInfo, advertisedWsUrl).then((endpoint) => {
```
mirpo, line 28: Maybe location is a part of `PeerInfo`? For discussion.

src/logic/Tracker.js
```diff
@@ -51,11 +52,13 @@ module.exports = class Tracker extends EventEmitter {
         const source = statusMessage.getSource()
         const status = statusMessage.getStatus()
         const { rtts } = status
+        const { location } = status
```
mirpo: `const { rtts, location } = status`

mirpo (changes requested):
LGFM

src/composition.js
juslesan: Yes, I was thinking about this possibility also. Maybe it's a better idea to add it there

src/composition.js
juslesan: It's not necessary here but could be used to add locations to the trackers in the future

mirpo: Let's remove it, if it's not in use now.

src/composition.js
mirpo: I think we should do it now.

bin/tracker.js
juslesan: Here tracker returns {"statusCode":500,"error":"Internal Server Error","message":"Node not found"} to clients

harbu (approved):

PR #445, opened 2020-08-12
geoip look up to tracker

juslesan:
Tracker looks up location based on ip given both of these conditions:

- The location data given by a node on status is not valid

- The location data stored in tracker is not valid


src/logic/Tracker.js
```diff
@@ -240,6 +241,37 @@ module.exports = class Tracker extends EventEmitter {
         return topology
     }
 
+    _updateLocation(node, location) {
+        if (this._isValidNodeLocation(location)) {
+            this.nodeLocations[node] = location
+        } else if (!this._isValidNodeLocation(this.getNodeLocation(node)) && !this._isValidNodeLocation(location)) {
```
harbu: I think you can leave out 2nd condition as it is implied by the use of `else if`.

src/logic/Tracker.js
```diff
@@ -240,6 +241,37 @@ module.exports = class Tracker extends EventEmitter {
...
+        const address = this.protocols.trackerServer.endpoint.peerBook.getAddress(node)
+        if (address) {
+            return geoiplite.lookup(address.split(':')[1].replace('//', ''))
+        }
+        return null
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    _isValidNodeLocation(location) {
+        return location && !(!location.country && !location.city && !location.latitude && !location.longitude)
```
harbu: Should be equivalent to
```js
location && (location.country || location.city || location.latitude || location.longitude)
```

src/logic/Tracker.js
```diff
@@ -240,6 +241,37 @@ module.exports = class Tracker extends EventEmitter {
...
+            if (geoip) {
+                this.nodeLocations[node] = {
+                    country: geoip.country,
+                    city: geoip.city,
+                    latitude: geoip.ll[0],
+                    longitude: geoip.ll[1]
+                }
+            }
+        } else {
+            this.nodeLocations[node] = null
```
harbu: So if the given `location` is not valid but the current location `this.getNodeLocation(node)` is, we set the current location to be `null`? Would it be better to keep the current location as is?

harbu (changes requested):
Maybe we could use an integration test for the scenario where a node reports `location: null` and then the tracker uses geo-ip to figure out location?

harbu (approved):

PR #452, opened 2020-09-17
promise timeouts for connects

juslesan:
- Add timeouts to promises for connecting nodes

### DIFF
--- a/src/helpers/PromiseTimeout.js
+++ b/src/helpers/PromiseTimeout.js
@@ -1,0 +1,18 @@
+const promiseTimeout = (ms, promise) => {
+    const timeout = new Promise((resolve, reject) => {
+        const id = setTimeout(() => {
+            clearTimeout(id)
+            // eslint-disable-next-line prefer-promise-reject-errors
+            reject('timed out in ' + ms + 'ms.')
+        }, ms)
+    })
+
+    return Promise.race([
+        promise,
+        timeout
+    ])
+}
+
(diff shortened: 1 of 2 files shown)


src/helpers/PromiseTimeout.js
```diff
@@ -0,0 +1,18 @@
+const promiseTimeout = (ms, promise) => {
+    const timeout = new Promise((resolve, reject) => {
+        const id = setTimeout(() => {
+            clearTimeout(id)
```
harbu: Since this callback is invoked upon the timeout firing, I don't think the clearing here is strictly necessary.

src/helpers/PromiseTimeout.js
```diff
@@ -0,0 +1,18 @@
...
+            // eslint-disable-next-line prefer-promise-reject-errors
+            reject('timed out in ' + ms + 'ms.')
+        }, ms)
+    })
+
+    return Promise.race([
+        promise,
```
harbu, line 15: One minor optimization: on `promise` resolve clear `timeout`.

harbu (approved):

src/helpers/PromiseTimeout.js
harbu: The idea being that if we call `promiseTimeout(SOME_BIG_NUMBERE_HERE, connect)` and `connect` resolves, then the timeout gets cleared immediately.

PR #454, opened 2020-09-18
Fixing connections

juslesan:
- Add logic for clearing up failed connections

- Timeout new connections after 10 seconds of no data channel being formed


harbu (approved):
LGTM; maybe re-try attempts and timeout as named constants in beginning of file?

PR #456, opened 2020-09-28
One tracker port and SSL support

mirpo:
- added SSL support 
- added test for all tracker endpoints (as we are going to use them for public service)
- added metrics endpoint for the tracker

breaking change:
`startTracker` now expects config object. if it looks good I can refactor `startNetworkNode` and `startStorageNode`

Fixed bug with string partition:
https://github.com/streamr-dev/network/blob/169118583eb49710a50216d3af0429bb7de81c26/src/logic/Tracker.js#L225-L226


src/logic/Tracker.js
```diff
@@ -222,8 +222,8 @@ module.exports = class Tracker extends EventEmitter {
             streamKeys = Object.keys(this.overlayPerStream).filter((streamKey) => streamKey.includes(streamId))
         } else {
             let askedStreamKey = null
-            if (streamId && partition) {
-                askedStreamKey = new StreamIdAndPartition(streamId, parseInt(partition, 10))
+            if (streamId && !Number.isNaN(partition) && partition >= 0) {
```
timoxley: Note `Number.isNaN(Infinity) // false`

You probably actually want: `Number.isSafeInteger(partition)`

src/identifiers.js
```diff
@@ -28,7 +28,7 @@ class StreamIdAndPartition {
 
     static fromKey(key) {
         const [id, partition] = key.split('::')
-        return new StreamIdAndPartition(id, Number.parseInt(partition, 10))
+        return new StreamIdAndPartition(id, parseInt(partition))
```
timoxley: This seems… worse? why

src/logic/Tracker.js
```diff
@@ -222,8 +222,8 @@ module.exports = class Tracker extends EventEmitter {
...
+                askedStreamKey = new StreamIdAndPartition(streamId, parseInt(partition))
```
timoxley: again why remove radix

test/benchmarks/overlay-topology-randomness.js
```diff
@@ -6,7 +6,7 @@ const numOfNodes = 1000
 const printProgress = true
 
 const idxToNodeId = (idx) => `${idx + 1}`
-const nodeIdToIdx = (nodeId) => parseInt(nodeId, 10) - 1
+const nodeIdToIdx = (nodeId) => parseInt(nodeId) - 1
```
timoxley: again why remove radix

.eslintrc.js
```diff
@@ -8,6 +8,8 @@ module.exports = {
         'plugin:promise/recommended'
     ],
     rules: {
+        'newline-per-chained-call': DISABLED,
+        radix: ['error', 'as-needed'],
```
timoxley: why remove radix

bin/tracker.js
```diff
@@ -20,12 +16,14 @@ program
     .option('--port <port>', 'port', 30300)
     .option('--ip <ip>', 'ip', '0.0.0.0')
     .option('--maxNeighborsPerNode <maxNeighborsPerNode>', 'maxNeighborsPerNode', 4)
+    .option('--exposeHttpEndpoints <exposeHttpEndpoints>', 'exposeHttpEndpoints', true)
```
timoxley: Maybe could do with better description, seems weird that a flag defaulting to a boolean would take a parameter

bin/tracker.js
```diff
@@ -51,84 +49,32 @@ if (program.sentryDns) {
...
-        logger.info(`tracker is listening on ${address}`)
-    })
-}
+async function main() {
+    try {
+        const tracker = await startTracker({
+            host: program.ip,
+            port: parseInt(program.port),
+            id,
+            maxNeighborsPerNode: parseInt(program.maxNeighborsPerNode),
```
timoxley: why no radix

bin/tracker.js
```diff
@@ -51,84 +49,32 @@ if (program.sentryDns) {
...
+            privateKeyFileName: program.privateKeyFileName,
+            certFileName: program.certFileName
+        })
 
-startTracker(program.ip, parseInt(program.port, 10), id, parseInt(program.maxNeighborsPerNode, 10), name)
-    .then((tracker) => {
-        logger.info('started tracker id: %s, name: %s, port: %d, ip: %s, maxNeighborsPerNode: %d, '
-            + 'metrics: %s, metricsInterval: %d, apiKey: %s, streamId: %s, sentryDns: %s',
-        id, name, program.port, program.ip, program.maxNeighborsPerNode, program.metrics,
+        logger.info('started tracker id: %s, privateKeyFileName: %s, certFileName: %s\n, '
```
timoxley: Consider logging an object with `%o`, gives you auto-named fields formatted on a single line:

```js
logger.info('started tracker: &o', {
  id,
  name,
  ...program
})
```

bin/tracker.js
```diff
@@ -51,84 +49,32 @@ if (program.sentryDns) {
...
-        if (client || program.metrics) {
+        if (program.metrics && program.apiKey && program.streamId) {
+            const client = new StreamrClient({
+                auth: {
+                    apiKey: program.apiKey
+                },
+                autoConnect: false
+            })
             setInterval(async () => {
                 const metrics = await tracker.getMetrics()
```
timoxley, line 87: ooo  unrelated to this PR, but should never do async work inside a `setInterval`
If `await tracker.getMetrics()` ever takes longer than the interval, you can end up queuing endless work potentially crashing the process.

Additionally `setInterval` is bad. If the event loop gets blocked for a significant period, e.g. performing a massive GC or doing a debug memory snapshot, so you can end up running a large number of the `setInterval` callbacks in quick succession… which could also blow the process up by trying to do too much work in one go, or in this case could give you funky metrics by sending the same metrics object N times because all the queued `setInterval` callbacks executed at once. Additionally if there's queued calls and they happen to resolve out-of-order, then you're going to be sending your metrics results out-of-order, which could be misleading.

To ensure only one of these is ever happening at once, and callbacks execute at most once every `program.metricsInterval`, this should be a recursive function with a `setTimeout`, something like:

```js
;(async function pollMetrics() {
    const metrics = await tracker.getMetrics()
    // send metrics to streamr.network
    if (client) {
        client.publishHttp(program.streamId, metrics)
    }

    // output to console
    if (program.metrics) {
        logger.info(JSON.stringify(metrics, null, 3))
    }
    clearTimeout(pollMetrics.t) // always clear timeout before you set it
    pollMetrics.t = setTimeout(pollMetrics, program.metricsInterval)
}())
```

bin/tracker.js
```diff
@@ -139,21 +85,25 @@ startTracker(program.ip, parseInt(program.port, 10), id, parseInt(program.maxNei
 
                 // output to console
                 if (program.metrics) {
-                    logger.log(JSON.stringify(metrics, null, 3))
+                    logger.info(JSON.stringify(metrics, null, 3))
```
timoxley: 3 spaces :/

bin/tracker.js
```diff
@@ -139,21 +85,25 @@ startTracker(program.ip, parseInt(program.port, 10), id, parseInt(program.maxNei
...
         }
-
-        if (program.endpointServerPort) {
-            startServer(tracker, program.endpointServerPort)
-        }
-    })
-    .catch((err) => {
+    } catch (err) {
         logger.error(err)
         process.exit(1)
```
timoxley, line 102: Also unrelated, but this should also probably be `pino.final(logger).error(err)` otherwise the process can exit before the buffer is flushed.

https://github.com/pinojs/pino/blob/db651a51762e8b043f406a73bb19470bcf6dcff1/docs/asynchronous.md#log-loss-prevention

bin/tracker.js
```diff
@@ -51,84 +49,32 @@ if (program.sentryDns) {
...
-            throw err
-        }
-        logger.info(`tracker is listening on ${address}`)
-    })
-}
+async function main() {
+    try {
+        const tracker = await startTracker({
+            host: program.ip,
+            port: parseInt(program.port),
```
timoxley: why no radix

bin/run.js
```diff
@@ -28,12 +27,11 @@ for (let i = 0; i < program.streams; i++) {
 let debug = false
 
 const productionEnv = Object.create(process.env)
-productionEnv.DEBUG = 'streamr:*,-streamr:connection:*'
-productionEnv.checkUncaughtException = true
+productionEnv.LOG_LEVEL = 'debug'
```
timoxley: I wonder if this should assign a default rather than forcibly overriding whatever was set in `env`?
```js
productionEnv.LOG_LEVEL = productionEnv.LOG_LEVEL || 'debug'
```

src/composition.js
```diff
@@ -7,22 +7,107 @@ const NodeToNode = require('./protocol/NodeToNode')
...
+        extraLogger.debug('request to /topology/')
+
+        res.end(JSON.stringify(tracker.getTopology()))
+    }).get('/topology/:streamId/', (res, req) => {
+        writeCorsHeaders(res, req)
+
+        const streamId = req.getParameter(0)
+        if (streamId === '') {
+            extraLogger.error('500 streamId must be a not empty string')
+            res.writeStatus('500 streamId must be a not empty string').end()
```
timoxley: This should probably `return` otherwise it will still run the code below

src/composition.js
```diff
@@ -7,22 +7,107 @@ const NodeToNode = require('./protocol/NodeToNode')
...
+
+        extraLogger.debug(`request to /topology/${streamId}/`)
+        res.end(JSON.stringify(tracker.getTopology(streamId, null)))
+    }).get('/topology/:streamId/:partition/', (res, req) => {
+        writeCorsHeaders(res, req)
+
+        const streamId = req.getParameter(0)
+        if (streamId === '') {
+            extraLogger.error('500 streamId must be a not empty string')
+            res.writeStatus('500 streamId must be a not empty string').end()
```
timoxley: `return`

src/composition.js
```diff
@@ -7,22 +7,107 @@ const NodeToNode = require('./protocol/NodeToNode')
...
+        }
+
+        const askedPartition = parseInt(req.getParameter(1))
+        if (Number.isNaN(askedPartition) || askedPartition < 0) {
+            extraLogger.error('500 partition must be a positive integer')
+            res.writeStatus('500 partition must be a positive integer').end()
```
timoxley: `return`

src/composition.js
```diff
@@ -7,22 +7,107 @@ const NodeToNode = require('./protocol/NodeToNode')
...
+        writeCorsHeaders(res, req)
+
+        const streamId = req.getParameter(0)
+        if (streamId === '') {
+            extraLogger.error('500 streamId must be a not empty string')
+            res.writeStatus('500 streamId must be a not empty string').end()
+        }
+
+        const askedPartition = parseInt(req.getParameter(1))
+        if (Number.isNaN(askedPartition) || askedPartition < 0) {
```
timoxley: `!Number.isSafeInteger(askedPartition) || askedPartition < 0`

Also consider logging the erroneous partition number in the error message.

src/composition.js
```diff
@@ -7,22 +7,107 @@ const NodeToNode = require('./protocol/NodeToNode')
...
+            extraLogger.error('500 partition must be a positive integer')
+            res.writeStatus('500 partition must be a positive integer').end()
+        }
+
+        extraLogger.debug(`request to /topology/${streamId}/${askedPartition}/`)
```
timoxley: I feel like all these logger messages should go into some kind of generic middleware, or/and should happen before any other calls so it always logs messages before potentially throwing.

This uWebsockets thing is weird. "Commercially supported" seems to actually mean "lack of examples or documentation", but maybe something like:


```js
wws.any((res, req) => {
    // http logger
    extraLogger.debug('request to %s: %s', req.getMethod(), req.getUrl())

    // attach json helper
    res.json = (...args) => {
        res.end(JSON.stringify(...args))
    }

    req.setYield(true) // make this middleware
}
```

src/composition.js
timoxley: `500` perhaps should be `422`, leave 5xx for unexpected errors and crashes?

timoxley (commented):
The radix removal seems worse to me? Some nitpicking and related/unrelated suggestions, but otherwise lgtm.

juslesan (approved):

src/logic/Tracker.js
mirpo: Replaced to isSafeInteger, but why would I care about `Infinity`?

src/identifiers.js
mirpo: reverted radix everywhere

bin/tracker.js
mirpo: Good point, changed to `flag`

bin/tracker.js
mirpo: Can't do this, it outputs the whole `commander` object with all internals. Changed a bit.

bin/tracker.js
mirpo: Ok, good to know. Let's improve it in separate PR.

bin/tracker.js
mirpo: Yeah, more readable

bin/tracker.js
mirpo: I took from official documentation.
https://getpino.io/#/docs/help?id=exit-logging#/docs/help?id=exit-logging#/docs/help?id=exit-logging

mirpo: Added

bin/run.js
mirpo: Good point

src/composition.js
mirpo: Yeah, just copy-pasta from `fast` implementation

src/composition.js
mirpo: Need to test separately

package.json
```diff
@@ -70,6 +70,7 @@
     "eslint-plugin-import": "^2.22.0",
     "eslint-plugin-import-order": "2.1.4",
     "eslint-plugin-promise": "^4.2.1",
+    "got": "^11.7.0",
```
harbu, line 73: I would prefer we use `node-fetch`. Not because find it to be best library, but to be consistent with other projects (such as broker).

bin/tracker.js
harbu: He is probably referring to odd number. Not 2 or 4 😄

src/composition.js
```diff
@@ -7,22 +7,110 @@ const NodeToNode = require('./protocol/NodeToNode')
...
-function startTracker(host, port, id = uuidv4(), maxNeighborsPerNode = 4, advertisedWsUrl = null, name, location, pingInterval) {
-    const peerInfo = PeerInfo.newTracker(id, name, location)
-    return startEndpoint(host, port, peerInfo, advertisedWsUrl, pingInterval).then((endpoint) => {
-        const opts = {
-            peerInfo,
-            protocols: {
-                trackerServer: new TrackerServer(endpoint)
-            },
-            maxNeighborsPerNode
+const writeCorsHeaders = (res, req) => {
```
harbu: All this HTTP stuff needs to be moved to a separate file. This file is about composing functionality for end-user needs and handling configuring.

src/composition.js
```diff
@@ -7,22 +7,110 @@ const NodeToNode = require('./protocol/NodeToNode')
...
+    wss.get('/topology/', (res, req) => {
+        writeCorsHeaders(res, req)
+        extraLogger.debug('request to /topology/')
+
+        res.end(JSON.stringify(tracker.getTopology()))
+    }).get('/topology/:streamId/', (res, req) => {
+        writeCorsHeaders(res, req)
+
+        const streamId = req.getParameter(0)
+        if (streamId === '') {
```
harbu: Not sure we would every hit this with the mapping `wss.get('/topology/', ...)` above though?

src/composition.js
```diff
@@ -7,22 +7,110 @@ const NodeToNode = require('./protocol/NodeToNode')
...
+            extraLogger.error('422 streamId must be a not empty string')
+            res.writeStatus('422 streamId must be a not empty string').end()
+            return
+        }
+
+        const askedPartition = Number.parseInt(req.getParameter(1), 10)
+        if (!Number.isSafeInteger(askedPartition) || askedPartition < 0) {
+            extraLogger.error(`422 partition must be a positive integer, askedPartition: ${askedPartition}`)
```
harbu: From the server-side's viewpoint it is not an error for a user to send an invalid partition. The server is doing exactly what is supposed to: rejecting the request. It is a user request error. So IMO we shouldn't log an error.

src/composition.js
```diff
@@ -7,22 +7,110 @@ const NodeToNode = require('./protocol/NodeToNode')
...
-        return new Tracker(opts)
     })
 }
 
+const startTracker = async ({
+    host, port, id = uuidv4(), exposeHttpEndpoints = true,
+    maxNeighborsPerNode = 4, advertisedWsUrl = null, name, location, pingInterval,
+    privateKeyFileName, certFileName
+}) => {
+    const [wss, listenSocket] = await startWebSocketServer(host, port, privateKeyFileName, certFileName)
```
harbu: To be consistent with rest of functions I guess you could keep
```js
const ep = startEndpoint(host, port, peerInfo, advertisedWsUrl, pingInterval)
```
as in other functions

and then attach to HTTP endpoints
```js
trackerHttpEndpoints(endpoint.wss, tracker)
```

test/integration/tracker-endpoints.test.js
```diff
@@ -0,0 +1,100 @@
...
+
+        nodeOne.subscribe(streamId, 0)
+        nodeTwo.subscribe(streamId, 0)
+
+        nodeOne.subscribe(streamId2, 0)
+
+        nodeOne.addBootstrapTracker(tracker.getAddress())
+        nodeTwo.addBootstrapTracker(tracker.getAddress())
+
+        await wait(1000)
```
harbu: Replace sleep with event awaiting or waitForCondition for quicker tests.

harbu (changes requested):

bin/tracker.js
mirpo: 🍖

src/composition.js
mirpo: nope

harbu: So I guess my question would then be, why have code that is never run?

src/composition.js
harbu: You didn't address this.

harbu (approved):

test/integration/tracker-endpoints.test.js
```diff
@@ -0,0 +1,100 @@
...
+            expect(jsonResult['stream-2::0']).toBeUndefined()
+            done()
+        } catch (error) {
+            done(error)
+        }
+    })
+
+    it('/location/', async (done) => {
+        try {
+            const jsonResult = await got(`http://${LOCALHOST}:${trackerPort}/topology/stream-1/0/`).json()
```
timoxley: actually… why try/catch around all these? And why use `done()` at all?

timoxley (approved):

PR #462, opened 2020-10-14
Make startNetworkNode, startStorageNode take objects instead of argument lists

harbu:
- _Breaking change_: Functions `startNetworkNode` and `startStorageNode` now take a single object argument instead of a list of arguments. This is (a) to be consistent with `startTracker` and also (b)  the number of positional arguments was getting quite unruly.
```js
// before
n1 = await startNetworkNode('127.0.0.1', 30411, 'node-1')

// after
n1 = await startNetworkNode({
    host: '127.0.0.1',
    port: 30411,
    id: 'node-1',
    trackers: ['ws://tracker-address:30300']
})
```

- _Breaking change_: Method `Node#addBoostrapTracker()` is removed. List of trackers is now given when a Node is created (e.g. with `startNetworkNode`)
- _Breaking change_: Method `Node#start()` added to make an instance of `Node` connect to its trackers. Before this is called, the Node will not yet do anything.
- _Breaking change_: Option `exposeHttpEndpoints` is now called `attachHttpEndpoints`.

### DIFF
--- a/bin/tracker.js
+++ b/bin/tracker.js
@@ -16,7 +16,7 @@ program
     .option('--port <port>', 'port', 30300)
     .option('--ip <ip>', 'ip', '0.0.0.0')
     .option('--maxNeighborsPerNode <maxNeighborsPerNode>', 'maxNeighborsPerNode', 4)
-    .option('--exposeHttpEndpoints', 'expose http endpoints')
+    .option('--attachHttpEndpoints', 'attach http endpoints')
     .option('--apiKey <apiKey>', 'apiKey for StreamrClient', undefined)
     .option('--streamId <streamId>', 'streamId for StreamrClient', undefined)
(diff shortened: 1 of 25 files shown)


juslesan (approved):

timoxley:
@harbu any idea about the test failures?

harbu:
> @harbu any idea about the test failures?

Fixed it. We have a hashing method that will take a streamId and then assign it consistently to one of the (three, in this test) trackers. I changed the method so the assignments changed a little as well. Admittedly this is rather arbitrary, as I noted in the test comment:
https://github.com/streamr-dev/network/pull/462/commits/07aefe7a1325ac9b03409c899e58a5541b2b9a01#diff-c3377040339c13a74c83c0604e5037bcf1654e8f9ff76b2e0a8329361793d88dR8-R11

PR #463, opened 2020-10-15
Remove unused code & prepare for resend pull change

harbu:
The inherit problem of resends seems to be the fact that they are push-based instead of pull-based. In preparation of changing the logic to adhere to the pull-pattern, I did some clean up in this PR.

- Remove unused class `MemoryStorage`
- Remove `AskNeighborsResendStrategy` (L2 strategy), we are dropping support for this feature and just keeping L1 & L3 around.
- Rename `StorageResendStrategy` to `LocalResendStrategy`
- Rename `StorageNodesResendStrategy` to `ForeignResendStrategy`
- Move all resend related logic to its own folder


juslesan (approved):
LGTM

timoxley (approved):
👍

PR #464, opened 2020-10-19
Paused streams

harbu:


harbu:
The issue turned out not to be that the streams were flowing instead of paused. Got buffering working correctly with flowing streams as well, so closing this as it is no longer needed. Might have value in future though if we decide to go this route.

PR #469, opened 2020-10-26
refactor metrics

harbu:
- Include and unify all metrics related functionality in `MetricsContext.js`
- MetricsContext passed via constructors down.


juslesan (approved):
LGTM

PR #477, opened 2020-12-01
NET-91 (Latency) Metrics improvements

bearni95:


test/integration/latency.test.js
```diff
@@ -0,0 +1,99 @@
...
+
+    beforeEach(async () => {
+        tracker = await startTracker({
+            host: '127.0.0.1',
+            port: trackerPort,
+            id: 'tracker'
+        })
+        // disable trackers formAndSendInstructions function
+        // eslint-disable-next-line no-underscore-dangle
+        tracker._formAndSendInstructions = () => {}
```
harbu: Is this disabling necessary? Can be probably removed.

test/integration/latency.test.js
```diff
@@ -0,0 +1,99 @@
...
+        node1 = await startNetworkNode({
+            host: '127.0.0.1',
+            port: port1,
+            id: 'node1',
+            trackers: [tracker.getAddress()]
+        })
+
+        node2 = await startNetworkNode({
```
harbu: `node2` not used in this test. Can be removed?

test/integration/latency.test.js
```diff
@@ -0,0 +1,99 @@
+const { StreamMessage, MessageID, MessageRef } = require('streamr-client-protocol').MessageLayer
+
+const { startNetworkNode, startTracker } = require('../../src/composition')
+
+describe('TestLatency', () => {
```
harbu: It is convention to use camel case when we are referring to a specific class. Here we don't have a class, so maybe a free form description is more appropriate, e.g. `latency metrics`.

test/integration/latency.test.js
```diff
@@ -0,0 +1,99 @@
...
+    const trackerPort = 32907
+
+    let node1
+    const port1 = 33978
+
+    let node2
+    const port2 = 33979
+
+    beforeEach(async () => {
+        tracker = await startTracker({
```
harbu, line 13: You may not need `tracker` in this test actually with only one node.

test/integration/latency.test.js
```diff
@@ -0,0 +1,99 @@
...
+        await node1.stop()
+        await tracker.stop()
+    })
+
+    it('TestLatency should fetch empty metrics', async () => {
+        const metrics = await node1.metrics.report()
+        expect(metrics.latency.last).toEqual(0)
+    })
+
+    it('Should send a single message to Node1 and collect latency', (done) => {
```
harbu: start description with lowercase to be consistent with other tests

test/integration/latency.test.js
```diff
@@ -0,0 +1,99 @@
...
+                'msgChainId'
+            ),
+            prevMsgRef: new MessageRef(0, 0),
+            content: {
+                messageNo: 1
+            },
+        }))
+    })
+
+    it('Should send a bunch of messages to Node1 and collect latency', async (done) => {
```
harbu: start description with lowercase to be consistent with other tests

test/integration/latency.test.js
```diff
@@ -0,0 +1,99 @@
...
+
+        node2.start()
+    })
+
+    afterEach(async () => {
+        await node1.stop()
+        await tracker.stop()
+    })
+
+    it('TestLatency should fetch empty metrics', async () => {
```
harbu: start description with lowercase to be consistent with other tests

harbu (changes requested):

test/integration/latency.test.js
bearni95: The node initialization throws an error when started without a tracker, seems to need it

harbu (approved):

PR #478, opened 2020-12-01
NET-97: Tracker StorageNodesResponse filtering

harbu:
If a storage node itself requests storage nodes from tracker, the requesting node should not be itself included in the response. Wrote a test for this and fixed it.


juslesan (approved):
LGTM

PR #480, opened 2020-12-09
NET-21: TypeScript

harbu:
Convert all source files into TypeScript, set up TypeScript for IntelliJ, Jest, as well as building project sources into JavaScript distributable with type definitions.

## Notable new files and directories
- tsconfig.json - Settings for TypeScript.
- dist/ - Compiled JS will be built into this directoy along with type declarations of form `*.d.ts`.
- src/types - Self declared type declarations for libraries that don't provide their own typings.

## Notable changes to existing files
- jest.config.js - Configured to work with TypeScript now. You could start writing tests in TypeScript.
- package.json - Added `build` and `prepare` tasks for building project sources.
- .gitignore - Ignore compiled JS in _dist/_ folder
- .npmignore - Ignore src/ and test/ folders from built package. Overall be more picky about what gets included in npm package.

## Next steps and limitations
[ ] <del>Get this running green in GitHub Actions.</del>
[ ] <del>Test out that produced NPM package actually works.</del>
- Tests have not been converted to TypeScript. Before converting tests to TypeScript, good idea to get the `streamr-client-protocol` library to use TypeScript first (or at the very least, produce associated typings in a d.ts file.
- Certain types could be even more strictly or elegantly defined. I'd expect such improvements to be made in the future, rather than fixate on them now. Unless of course something is totally broken...
- Linting for TypeScript. Look into what is out there, do we even necessarily need?


juslesan (approved):
LGTM, the bin files should be updated to work with the new ts files

harbu:
The bin scripts have been fixed now. Had to unfortunately keep them as JS files because converting them to TS somehow caused the compiler to totally change the output directory structure which breaks everything.

PR #482, opened 2020-12-11
NET-116: Add tracker endpoint /topology-union/

teogeb:
Returns all known connections between nodes, across all streams. Cache the result for 15 seconds.

### DIFF
--- a/src/helpers/trackerHttpEndpoints.js
+++ b/src/helpers/trackerHttpEndpoints.js
@@ -15,17 +15,20 @@ const respondWithError = (res, req, errorMessage) => {
 }
 
 const trackerHttpEndpoints = (wss, tracker, metricsContext) => {
-    wss.cachedJsonGet = (endpoint, maxAge, jsonFactory) => {
-        let cache = undefined
+    wss.cachedJsonGet = (endpoint, maxAge, jsonFactory) => { // eslint-disable-line no-param-reassign
+        let cache
         return wss.get(endpoint, (res, req) => {
             extraLogger.debug('request to ' + endpoint)
             writeCorsHeaders(res, req)
             if ((cache === undefined) || (Date.now() > (cache.timestamp + maxAge))) {
-                cache = { json: jsonFactory(), timestamp: Date.now() }
+                cache = {
+                    json: jsonFactory(),
+                    timestamp: Date.now()
+                }
             }
             res.end(JSON.stringify(cache.json))
         })
-    };
+    }
     wss.get('/topology/', (res, req) => {
         extraLogger.debug('request to /topology/')
         writeCorsHeaders(res, req)
--- a/src/helpers/trackerHttpEndpoints.js
+++ b/src/helpers/trackerHttpEndpoints.js
@@ -1,5 +1,7 @@
 const _ = require('lodash')
 
+const { getTopology, getTopologyUnion } = require('../logic/TopologyFactory')
+
 const extraLogger = require('./logger')('streamr:tracker:http-endpoints')
 
 const writeCorsHeaders = (res, req) => {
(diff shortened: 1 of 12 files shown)


src/logic/Tracker.js
```diff
@@ -217,6 +217,19 @@ module.exports = class Tracker extends EventEmitter {
...
+    getTopologyUnion() {
+        const mergeSetMapInto = (target, source) => {  // merges each source value (a Set object) into the target value with the same key
+            source.forEach((sourceSet, key) => {
+                const targetSet = target.get(key)
+                const mergedSet = (targetSet !== undefined) ? new Set([...targetSet, ...sourceSet]) : sourceSet
+                target.set(key, mergedSet)
+            })
+            return target
+        }
+        const nodeMaps = Object.values(this.overlayPerStream).map(topology => new Map(Object.entries(topology.nodes)))
```
harbu: As `topology.nodes` could be thought of as a private member of `OverlayTopology`, I feel a bit uneasy of accessing it directly from other classes (at least without TypeScript, in which we can explicitly declare a field public). Could `OverlayTopology#state()` be used here? Or alternatively, creating a getter method on OverlayTopology such as `OverlayTopology#getNodes`.

src/helpers/trackerHttpEndpoints.js
```diff
@@ -15,6 +15,20 @@ const respondWithError = (res, req, errorMessage) => {
 }
 
 const trackerHttpEndpoints = (wss, tracker, metricsContext) => {
+    wss.cachedJsonGet = (endpoint, maxAge, jsonFactory) => { // eslint-disable-line no-param-reassign
```
harbu, line 22: Hmm.. don't see param re-assignment though? ``// // eslint-disable-line no-param-reassign``

src/helpers/trackerHttpEndpoints.js
```diff
@@ -48,6 +62,9 @@ const trackerHttpEndpoints = (wss, tracker, metricsContext) => {
         extraLogger.debug(`request to /topology/${streamId}/${askedPartition}/`)
         writeCorsHeaders(res, req)
         res.end(JSON.stringify(tracker.getTopology(streamId, askedPartition)))
+    }).cachedJsonGet('/topology-union/', 15 * 1000, () => {
+        const topologyUnion = tracker.getTopologyUnion()
+        return Object.fromEntries(new Map(Array.from(topologyUnion, ([key, value]) => [key, Array.from(value)])))
```
harbu: Wondering if these `Map <-> Array <-> Object` conversions really necessary or could be simplified?

harbu (changes requested):

kare:
> NET-115 Add dependency to org.jetbrains:annotations

@teogeb This is not NET-115 ticket!

src/helpers/trackerHttpEndpoints.js
teogeb: ESlint is complaining about the `wss` parameter of trackerHttpEndpoints function, which we modify:
`Assignment to property of function parameter 'wss'  no-param-reassign`

There is no optimal place for initializing this kind of middleware method. Maybe it is ok inside `trackerHttpEndpoints` as that utility method is specifically for HTTP requests.

src/logic/Tracker.js
teogeb: Created getter `OverlayTopology#getNodes`.

src/helpers/trackerHttpEndpoints.js
teogeb: Simplified the logic: now the intermediate state is just a plain object, not Map. The data structure is always so small that it doesn't make any performance difference. 

Added dependency to `lodash` library. Currently only the `mapValues` function is used, but I guess we'll use other functions in the near future, too. When we use "standard" lodash functions (instead of our own implementations), the code is more readable as many developers already know those basic functions and their behaviour.

harbu (approved):

PR #483, opened 2020-12-14
rework instructionThrottler

juslesan:
- Rework instruction handler to allow some parallelism and remove locking


src/logic/InstructionThrottler.js
```diff
@@ -1,20 +1,18 @@
-const { cancelable } = require('cancelable-promise')
-
 const logger = require('../helpers/logger')('streamr:InstructionThrottler')
```
harbu, line 1: Since `cancelable-promise` package is only used here, it should be cleaned up from package.json and package-lock.json.

harbu (approved):
LGTM; see comment about removing dependency

PR #485, opened 2020-12-17
Rework examples folder

harbu:
I removed all the old examples as they were unmaintained. I added a new toy example of using the network in a pub/sub setting where cpu/memory/system metrics is being published by one node and two other nodes subscribe to this data. The dependency is now directly to the `streamr-network` package which means that the example is not directly linked to changes in the project sources. With that being said, we should take care to every now and then update the example to work on a newer version of `streamr-network`.


fonty1:
Reworked it a little @harbu to allow for some demo flair, i.e. Having the Pub and Sub in separate terminals.

harbu:
LGTM!

PR #492, opened 2020-12-22
NET-140: messageQueue set max limit & other fixes

harbu:
- Extract MessageQueue as its own class to a new file
- MessageQueue has max limit
- Refactor `attemptToFlushMessages` logic
- Ensure `attemptToFlushMessages` increments send retries even if connection not opened
- Ability to set `flushRetryTimeout` for re-trying flushing


juslesan (approved):
LGTM

PR #493, opened 2020-12-22
NET-140: fix issue in production

harbu:
Will have to go thru these changes to see what is worth keeping around. The most important fix is 
beaf852 which seems to fix the issue we were seeing in production (dead connection was kept around forever).

Changes introduced:
- Make sure `PEER_DISCONNECTED` is emitted only once when `Connection#onError` is invoked.
- Remove connection retry logic from ping-pong.
- Send status to Tracker when node disconnects (important)

### DIFF
--- a/src/connection/WebRtcEndpoint.ts
+++ b/src/connection/WebRtcEndpoint.ts
@@ -181,7 +181,6 @@ export class WebRtcEndpoint extends EventEmitter {
                 delete this.connections[targetPeerId]
             },
             onError: (err) => {
-                this.emit(Event.PEER_DISCONNECTED, connection.getPeerInfo())
                 this.emit(`errored:${connection.getPeerId()}`, err)
             },
             onBufferLow: () => {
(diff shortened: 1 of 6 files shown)


src/connection/WebRtcEndpoint.ts
```diff
@@ -57,8 +57,8 @@ export class WebRtcEndpoint extends EventEmitter {
         stunUrls: string[],
         rtcSignaller: RtcSignaller,
         metricsContext: MetricsContext,
-        pingIntervalInMs = 5 * 1000,
-        newConnectionTimeout = 5000
+        pingIntervalInMs = 30 * 1000,
```
juslesan: Need to keep in mind that the default values here are set quite high

juslesan (approved):
LGTM

src/connection/WebRtcEndpoint.ts
harbu: Reverted these timeouts back

harbu:
Ended up reverting 4427abe and timeouts in 8729df6

PR #496, opened 2021-01-07
NET-124: Typescript tests

teogeb:
Converted tests to TypeScript. All tests pass, but some should be refactored later. 

Created separate `tsconfig.json` configs for development (e.g. running tests) and building the distribution package. With this setup the test code is excluded from the distribution package.

Added TypeScript support to eslint, but disabled some rules so that the tests pass the validation. In NET-125 we'll define a good lint rule set, and most likely we could re-enable some of the rules which we disabled here.

Suggested refactoring:

- Tests should not use private fields (NET-153)
- Refine the start logic of startNetworkNode/startStorageNode (NET-154).


.eslintrc.js
```diff
@@ -14,6 +16,22 @@ module.exports = {
         }],
         'max-classes-per-file': DISABLED,
         'promise/always-return': WARN,
-        'import/no-unresolved': DISABLED
+        'import/no-unresolved': DISABLED,
+        'import/extensions': DISABLED,
+        // TODO remove some of these is possible
```
harbu, line 21: What about this TODO and these ignores, seems like things to address since they come from `@typescript-eslint/recommended`?

package.json
```diff
@@ -61,6 +61,8 @@
     "@types/pino": "^6.3.4",
     "@types/uuid": "^8.3.0",
     "@types/ws": "^7.4.0",
+    "@typescript-eslint/eslint-plugin": "^4.12.0",
+    "@typescript-eslint/parser": "^4.12.0",
     "eslint": "^7.16.0",
     "eslint-config-airbnb-base": "^14.2.1",
     "eslint-config-streamr-nodejs": "^1.3.0",
```
harbu, line 68: Probably eslint packages here we can get rid of now then? Like 
```js
eslint-config-airbnb-base
eslint-config-streamr-nodejs
eslint-plugin-import
eslint-plugin-import-order
eslint-plugin-promise
```

test/integration/Connection.test.ts
```diff
@@ -1,8 +1,9 @@
-const { waitForCondition, wait } = require('streamr-test-utils')
+import { DescriptionType } from 'node-datachannel'
+import { waitForCondition, wait } from 'streamr-test-utils'
 
-const { Connection } = require('../../src/connection/Connection')
+import { Connection } from '../../src/connection/Connection'
 
-function onConnectPromise(functions) {
+function onConnectPromise(functions: any) {
```
harbu: `functions` could be typed here for type safety

test/integration/counter-filtering-in-tracker.test.ts
```diff
@@ -22,12 +23,12 @@ const formStatus = (counter1, counter2, nodes1, nodes2) => ({
             counter: counter2
         }
     }
-})
+} as any)
```
harbu: Should not have to use `any` here though as return type is `Status`?

test/integration/l3-resend-requests.test.ts
```diff
@@ -1,15 +1,17 @@
...
-const { startNetworkNode, startStorageNode, startTracker } = require('../../src/composition')
-const { Event: NodeEvent } = require('../../src/logic/Node')
+import { startNetworkNode, startStorageNode, startTracker } from '../../src/composition'
+import { Event as NodeEvent } from '../../src/logic/Node'
 
 const { ControlMessage } = ControlLayer
 const { StreamMessage, MessageID, MessageRef } = MessageLayer
 
-const typesOfStreamItems = async (stream) => {
+const typesOfStreamItems = async (stream: any) => {
```
harbu: stream could be typed to be `Readable` or `NodeJS.ReadableStream` or whatever the type is

test/integration/request-resend-from-uninvolved-node.test.ts
```diff
@@ -1,15 +1,17 @@
...
-const { startNetworkNode, startStorageNode, startTracker } = require('../../src/composition')
-const { Event: NodeEvent } = require('../../src/logic/Node')
+import { startNetworkNode, startStorageNode, startTracker } from '../../src/composition'
+import { Event as NodeEvent } from '../../src/logic/Node'
 
 const { ControlMessage } = ControlLayer
 const { StreamMessage, MessageID, MessageRef } = MessageLayer
 
-const typesOfStreamItems = async (stream) => {
+const typesOfStreamItems = async (stream: any) => {
```
harbu, line 12: same comment as before

test/unit/InstructionCounter.test.ts
```diff
@@ -22,7 +22,7 @@ describe('InstructionCounter', () => {
                 },
             }
         }
-        const filtered = instructionCounter.filterStatus(status, 'node')
+        const filtered = instructionCounter.filterStatus(status as any, 'node')
```
harbu: could variable `status` be typed in this test?

harbu (changes requested):
Let's remove unneeded packages + are all new the eslint disabled things necessary?

Rest of my comments about enhancing typing can be ignored and done later imo.

.eslintrc.js
teogeb: Yes, will be handled in NET-125 (see the PR description).

package.json
teogeb: If we keep using `streamr-nodejs` as our base rule set, we need `eslint-config-airbnb-base,` `eslint-plugin-import` and `eslint-plugin-promise` as those are peer dependencies to `streamr-nodejs` (https://github.com/streamr-dev/eslint-config-streamr/blob/master/node.js/package.json). We can remove `eslint-plugin-import-order` in NET-125.

test/integration/Connection.test.ts
teogeb: Fixed

test/integration/counter-filtering-in-tracker.test.ts
teogeb: `Status` interface has these field that that are not in the mock object:
- rtts: `Rtts`
- location: `Location`
- started: `string`

Changed the return type to `Partial<Status>` and added casts to places where we use the object.

test/integration/l3-resend-requests.test.ts
teogeb: Modified to `NodeJS.ReadableStream` and added cast when calling `waitForStreamToEnd` (which requires `Readable`)

test/unit/InstructionCounter.test.ts
teogeb: Same as in earlier comment: using Partial<Status> as some fields are missing.

harbu (approved):

PR #497, opened 2021-01-08
NET-136: Remove status timeout

juslesan:
- Bring back old instruction throttler
- Add filtering for new instructions from tracker based on counter to the instruction throttler
- Remove timeout for sending messages

1. These changes significantly improve the performance of the networks topology formation.
2. Some flaky tests seem to be fixed with this change
3. Excellent results with 16-2k node aws experiments


test/integration/tracker-node-reconnect-instructions.test.js
```diff
@@ -75,10 +75,9 @@ describe('Check tracker instructions to node', () => {
                 streamId,
                 streamPartition: 0,
                 nodeIds: [],
-                counter: 0
+                counter: 3
```
harbu, line 78: Motivation of this change?

harbu (approved):

test/integration/tracker-node-reconnect-instructions.test.js
juslesan: The test will fail if the counter is set below three. This happens because the instructionThrottler will filter any counter values that it has seen already

PR #499, opened 2021-01-15
NET-125: ESLint rules

teogeb:
Fine-tuned the rule set: added TypeScript specific rules and removed redundant rules. Some rules are used directly instead of extending streamr-nodejs base rule set.


src/connection/WebRtcEndpoint.ts
```diff
@@ -3,17 +3,17 @@ import nodeDataChannel, { DescriptionType } from 'node-datachannel'
 import getLogger from '../helpers/logger'
 import { PeerInfo } from './PeerInfo'
 import { Connection } from './Connection'
-import { Metrics, MetricsContext } from "../helpers/MetricsContext";
+import { Metrics, MetricsContext } from '../helpers/MetricsContext';
```
harbu: Hmmm semicolon? Shouldn't linter fail?

src/helpers/SeenButNotPropagatedSet.ts
```diff
@@ -1,4 +1,4 @@
-import LRUCache from "lru-cache";
+import LRUCache from 'lru-cache';
```
harbu: Semicolon

harbu (approved):
LGTM; semicolons not getting detected?

src/connection/WebRtcEndpoint.ts
teogeb: Fixed

PR #505, opened 2021-01-25
NET-32: Storage node config

teogeb:
New required option `storageConfig` for starting a storage node. The option contains an interface, which we can query the streams that the storage node should store. We can also add a listener for the config to notify the storage node about configuration changes.

With this change the storage node stores only the configured streams, not all streams. It subscribes to the configured streams and doesn't need the tracker to orchestrate the subscriptions.


src/composition.ts
```diff
@@ -134,8 +139,17 @@ export function startNetworkNode(opts: NetworkNodeOptions): Promise<NetworkNode>
...
-    return startNode(opts, PeerInfo.newStorage)
+export async function startStorageNode(opts: StorageNodeOptions): Promise<NetworkNode> {
+    const node = await startNode(opts, PeerInfo.newStorage)
+    const storageConfig = opts.storageConfig
+    storageConfig.getStreams().forEach((stream) => {
+        node.subscribe(stream.id, stream.partition)
+    })
+    storageConfig.addChangeListener({
+        onStreamAdded: (stream: StreamIdAndPartition) => node.subscribe(stream.id, stream.partition),
+        onStreamRemoved: (stream: StreamIdAndPartition) => node.unsubscribe(stream.id, stream.partition)
```
harbu, line 150: There's a potential bug here if the broker is serving adapter clients for `stream.id` and `stream.partition`, well actually depends on how broker is implemented. Before invoking this callback the broker should check from `SubscriptionManager` (in that repo) that there are no clients being served the stream in question.

src/composition.ts
```diff
@@ -134,8 +139,17 @@ export function startNetworkNode(opts: NetworkNodeOptions): Promise<NetworkNode>
     return startNode(opts, PeerInfo.newNode)
 }
 
-export function startStorageNode(opts: NetworkNodeOptions): Promise<NetworkNode> {
-    return startNode(opts, PeerInfo.newStorage)
+export async function startStorageNode(opts: StorageNodeOptions): Promise<NetworkNode> {
+    const node = await startNode(opts, PeerInfo.newStorage)
+    const storageConfig = opts.storageConfig
+    storageConfig.getStreams().forEach((stream) => {
```
harbu, line 145: In a sense it almost feels like there is no reason for the network level to know about the details of how storage should react upon changes to stream lists. As in, this code could live entirely in broker. However, might be too early to make that judgment...

test/integration/multi-storages.test.ts
```diff
@@ -1,114 +0,0 @@
-import { Tracker } from '../../src/logic/Tracker'
-import { NetworkNode } from '../../src/NetworkNode'
-import { waitForEvent, waitForCondition } from 'streamr-test-utils'
-
-import { startStorageNode, startNetworkNode, startTracker } from '../../src/composition'
-import { Event as TrackerServerEvent } from '../../src/protocol/TrackerServer'
-import { Event as NodeEvent } from '../../src/logic/Node'
-import { getTopology } from '../../src/logic/trackerSummaryUtils'
-
-describe('multiple storage nodes', () => {
```
harbu, line 10: I'm guessing this test was removed since it revolves around tracker assigning storages to streams. Hmmm.. do we need some testing around multiple storages? Maybe only when the 2nd ticket is started.

test/integration/storage-config.test.ts
```diff
@@ -0,0 +1,101 @@
...
+import { startTracker, startNetworkNode, startStorageNode, Storage } from '../../src/composition'
+import { StreamIdAndPartition } from '../../src/identifiers'
+import { MockStorageConfig } from './MockStorageConfig'
+
+const { StreamMessage, MessageID } = MessageLayer
+
+const HOST = '127.0.0.1'
+
+const createMockStream = () => {
+    return new StreamIdAndPartition(uuidv4(), Math.floor(Math.random() * 100))
```
harbu, line 16: I wonder if a stream with a named id would be easier here for debug purposes, e.g. `storage-config-test-stream-1` ,  `storage-config-test-stream-2`,  `storage-config-test-stream-3` and so forth.

test/integration/storage-config.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        config.addStream(initialStream)
+        tracker = await startTracker({
+            host: HOST,
+            port: 49800,
+            id: 'tracker'
+        })
+        relayNode = await startNetworkNode({
+            host: HOST,
+            port: 49801,
+            id: 'relay',
```
harbu, line 56: I recommend having `id` match exactly the name of the variable in the code. Requires less mental mapping effort when reading logs and reflecting messages against code.

test/integration/storage-config.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    afterEach(async () => {
+        await tracker.stop()
+        await relayNode.stop()
+        await storageNode.stop()
+    })
+
+    it('initial stream', async () => {
+        const message = createMockMessage(initialStream)
+        relayNode.publish(message)
+        await waitForCondition(() => (storage.store as any).mock.calls.length > 0)
```
harbu, line 84: 👍

test/integration/storage-config.test.ts
```diff
@@ -0,0 +1,101 @@
...
+const createMockMessage = (stream: StreamIdAndPartition) => {
+    return new StreamMessage({
+        messageId: new MessageID(stream.id, stream.partition, Date.now(), 0, 'mock-publisherId', 'mock-msgChainId'),
+        content: {
+            'foo': 'bar'
+        }
+    })
+}
+
+const createStreamMessageMatcher = (message: MessageLayer.StreamMessage) => {
```
harbu, line 28: I'm guessing you've tested that this matcher can fail if the ids do not match?

test/integration/storage-config.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        config.addStream(stream)
+        const message = createMockMessage(stream)
+        relayNode.publish(message)
+        await waitForCondition(() => (storage.store as any).mock.calls.length > 0)
+        expect(storage.store).toHaveBeenCalledWith(createStreamMessageMatcher(message))
+    })
+
+    it('remove stream', async () => {
+        config.removeStream(initialStream)
+        await waitForCondition(() => (storageNode.getStreams().length === 0))
```
harbu, line 99: I guess there is no obvious way of testing that any published messages hereafter would not get stored?

src/composition.ts
harbu: Will create a separate ticket to address this issue in later part.

PR #510, opened 2021-02-01
Per stream instruction throttler

juslesan:
- Change bin/network.js and bin/subscriber.js streams flag to subscribe all nodes to n amount of streams
- Rework InstructionThrottler to work on per stream basis. Solves churn issues with multi stream throttling

### DIFF
--- a/src/logic/InstructionThrottler.ts
+++ b/src/logic/InstructionThrottler.ts
@@ -2,7 +2,6 @@ import { cancelable, CancelablePromiseType } from 'cancelable-promise'
 import { StreamIdAndPartition, StreamKey } from '../identifiers'
 import { TrackerLayer } from 'streamr-client-protocol'
 import getLogger from '../helpers/logger'
-import { startTracker } from "../composition"
 
 const logger = getLogger('streamr:logic:InstructionThrottler')
 
--- a/src/logic/InstructionThrottler.ts
+++ b/src/logic/InstructionThrottler.ts
@@ -36,7 +36,7 @@ export class InstructionThrottler {
 
     add(instructionMessage: TrackerLayer.InstructionMessage, trackerId: string): void {
         const streamId = StreamIdAndPartition.fromMessage(instructionMessage).key()
-        if (!this.instructionCounter[streamId] || this.instructionCounter[streamId]) {
+        if (!this.instructionCounter[streamId] || this.instructionCounter[streamId] <= instructionMessage.counter) {
             this.instructionCounter[streamId] = instructionMessage.counter
             this.queue[StreamIdAndPartition.fromMessage(instructionMessage).key()] = {
                 instructionMessage,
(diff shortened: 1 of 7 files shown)


src/logic/InstructionThrottler.ts
```diff
@@ -22,23 +22,35 @@ export class InstructionThrottler {
...
 
     constructor(handleFn: (instructionMessage: TrackerLayer.InstructionMessage, trackerId: string) => Promise<void>) {
         this.handleFn = handleFn
+        this.ongoingPromises = {}
     }
 
     add(instructionMessage: TrackerLayer.InstructionMessage, trackerId: string): void {
         const streamId = StreamIdAndPartition.fromMessage(instructionMessage).key()
-        if (!this.instructionCounter[streamId] || this.instructionCounter[streamId] <= instructionMessage.counter) {
+        if (!this.instructionCounter[streamId] || this.instructionCounter[streamId]) {
```
harbu: This condition looks wrong?

src/logic/InstructionThrottler.ts
```diff
@@ -48,40 +60,53 @@ export class InstructionThrottler {
...
+        delete this.ongoingPromises[streamId]
     }
 
     isIdle(): boolean {
-        return !this.handling
+        let idle = true
+        Object.values(this.ongoingPromises).forEach((promises) => {
+            if (promises.handling) {
+                idle = false
+                return
```
harbu: The `return` here will not short circuit the forEach-loop, so it will go through the entire list anyways even if a `true` is found earlier since the `return` applies to each instance of the inner function. To short circuit the search use traditional for loop or `find` or `some` e.g.
 ```js
return Object.values(this.ongoingPromises).some((p) => p.handling))
```

File: test/unit/InstructionThrottler.test.ts [hunk 3]
```diff
@@ -53,6 +53,7 @@ describe('InstructionThrottler', () => {
             [createInstruction('stream-1', 1), 'tracker-1'],
             [createInstruction('stream-1', 5), 'tracker-1']
         ])
+        console.log()
```
harbu: rm

test/integration/tracker-node-reconnect-instructions.test.ts
```diff
@@ -82,18 +80,15 @@ describe('Check tracker instructions to node', () => {
                 counter: 3
             }).serialize()
         )
-        // @ts-expect-error private field
-        await waitForEvent(nodeOne.trackerNode, TrackerNodeEvent.TRACKER_INSTRUCTION_RECEIVED)
-        await waitForEvent(nodeOne, NodeEvent.NODE_DISCONNECTED)
+        await waitForEvent(nodeOne, NodeEvent.NODE_UNSUBSCRIBED)
```
harbu, line 89: To check that a disconnection actually occurred as the title of the test case implies, I think it would be worth checking that
```js
expect(Object.keys(nodeTwo.nodeToNode.endpoint.connections).length).toBe(2)
```
holds before sending the instruction.

harbu (changes requested):

src/logic/InstructionThrottler.ts
juslesan: Yeah not sure how that got there

harbu (approved):

PR #511, opened 2021-02-02
Reattempt Instruction on interval

juslesan:
- Make fulfilling tracker instructions eventually consistent
- Forces status resends to the tracker, makes latency of topology eventually visible for all streams in the trackers, good for network-explorer

The InstructionRetryManager could be merged to the InstructionThrottler module to block retrying instructions if there is an ongoing instruction for a stream. #510

### DIFF
--- a/src/logic/InstructionRetryManager.ts
+++ b/src/logic/InstructionRetryManager.ts
@@ -1,0 +1,47 @@
+import getLogger from '../helpers/logger'
+import { StreamIdAndPartition, StreamKey } from "../identifiers"
+import { TrackerLayer } from "streamr-client-protocol"
+
+const logger = getLogger('streamr:logic:InstructionRetryManager')
+
+
+export class InstructionRetryManager {
+    private readonly handleFn: (instructionMessage: TrackerLayer.InstructionMessage, trackerId: string) => Promise<void>
+    private readonly intervalInMs: number
+    private instructionRetryIntervals: { [key: string]: NodeJS.Timeout }  // streamId => instructionMessage
+
+    constructor(handleFn: (instructionMessage: TrackerLayer.InstructionMessage, trackerId: string) => Promise<void>, intervalInMs: number) {
+        this.handleFn = handleFn
+        this.intervalInMs = intervalInMs || 20000
+        this.instructionRetryIntervals = {}
+    }
+
+    add(instructionMessage: TrackerLayer.InstructionMessage, trackerId: string): void {
+        const id = StreamIdAndPartition.fromMessage(instructionMessage).key()
+        if (this.instructionRetryIntervals) {
+            clearTimeout(this.instructionRetryIntervals[id])
+        }
+        this.instructionRetryIntervals[id] = setTimeout(() =>
+            this.retryFunction(instructionMessage, trackerId)
+        , this.intervalInMs)
+    }
+
+    retryFunction(instructionMessage: TrackerLayer.InstructionMessage, trackerId: string): void {
+        this.handleFn(instructionMessage, trackerId)
+        this.instructionRetryIntervals[StreamIdAndPartition.fromMessage(instructionMessage).key()] = setTimeout(() =>
+            this.retryFunction(instructionMessage, trackerId)
+        , this.intervalInMs)
+    }
+
+    removeStreamId(streamId: StreamKey): void {
+        clearTimeout(this.instructionRetryIntervals[streamId])
+        delete this.instructionRetryIntervals[streamId]
+    }
+
+    reset(): void {
+        Object.values(this.instructionRetryIntervals).forEach((timeout) => {
+            clearTimeout(timeout)
+        })
+        this.instructionRetryIntervals = {}
+    }
+}
--- a/src/logic/InstructionRetryManager.ts
+++ b/src/logic/InstructionRetryManager.ts
@@ -4,7 +4,6 @@ import { TrackerLayer } from "streamr-client-protocol"
 
 const logger = getLogger('streamr:logic:InstructionRetryManager')
 
-
 export class InstructionRetryManager {
     private readonly handleFn: (instructionMessage: TrackerLayer.InstructionMessage, trackerId: string) => Promise<void>
     private readonly intervalInMs: number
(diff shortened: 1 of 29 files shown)


src/logic/InstructionRetryManager.ts
```diff
@@ -0,0 +1,53 @@
+import getLogger from '../helpers/logger'
+import { StreamIdAndPartition, StreamKey } from "../identifiers"
+import { TrackerLayer } from "streamr-client-protocol"
+
+const logger = getLogger('streamr:logic:InstructionRetryManager')
+
+export class InstructionRetryManager {
+    private readonly handleFn: (instructionMessage: TrackerLayer.InstructionMessage, trackerId: string) => Promise<void>
+    private readonly intervalInMs: number
+    private instructionRetryIntervals: { [key: string]: NodeJS.Timeout }  // streamId => instructionMessage
```
harbu: comment out of date

src/logic/InstructionRetryManager.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    constructor(handleFn: (instructionMessage: TrackerLayer.InstructionMessage, trackerId: string) => Promise<void>, intervalInMs: number) {
+        this.handleFn = handleFn
+        this.intervalInMs = intervalInMs || 30000
+        this.instructionRetryIntervals = {}
+    }
+
+    add(instructionMessage: TrackerLayer.InstructionMessage, trackerId: string): void {
+        const id = StreamIdAndPartition.fromMessage(instructionMessage).key()
+        if (this.instructionRetryIntervals) {
```
harbu: `if (this.instructionRetryIntervals[id])`?

src/logic/InstructionRetryManager.ts
```diff
@@ -0,0 +1,53 @@
...
+            clearTimeout(this.instructionRetryIntervals[id])
+        }
+        this.instructionRetryIntervals[id] = setTimeout(() =>
+            this.retryFunction(instructionMessage, trackerId)
+        , this.intervalInMs)
```
harbu, line 25: Intentionally `this.intervalInMs` for 1st call? Or should it be `0`?

src/logic/InstructionRetryManager.ts
```diff
@@ -0,0 +1,53 @@
...
+const logger = getLogger('streamr:logic:InstructionRetryManager')
+
+export class InstructionRetryManager {
+    private readonly handleFn: (instructionMessage: TrackerLayer.InstructionMessage, trackerId: string) => Promise<void>
+    private readonly intervalInMs: number
+    private instructionRetryIntervals: { [key: string]: NodeJS.Timeout }  // streamId => instructionMessage
+
+    constructor(handleFn: (instructionMessage: TrackerLayer.InstructionMessage, trackerId: string) => Promise<void>, intervalInMs: number) {
+        this.handleFn = handleFn
+        this.intervalInMs = intervalInMs || 30000
```
harbu: Since parameter `intervalInMs` is defined to always be a number,  `|| 30000` can be left out. Unless the intention is to set it to `30000` if `0` is passed as argument value.

src/logic/Node.ts
```diff
@@ -127,6 +132,7 @@ export class Node extends EventEmitter {
         this.trackerRegistry = Utils.createTrackerRegistry(opts.trackers)
         this.trackerBook = {}
         this.instructionThrottler = new InstructionThrottler(this.handleTrackerInstruction.bind(this))
+        this.instructionRetryManager = new InstructionRetryManager(this.handleTrackerInstruction.bind(this), 10000)
```
harbu: Should have `this.instructionRetryInterval` as 2nd argument?

src/logic/Node.ts
```diff
@@ -209,6 +215,7 @@ export class Node extends EventEmitter {
         this.logger.debug('unsubscribeFromStream: remove %s from streams', streamId)
         this.streams.removeStream(streamId)
         this.instructionThrottler.removeStreamId(streamId.key())
+        this.instructionRetryManager.removeStreamId(streamId.key())
```
harbu, line 218: Should also add call to `this.instructionRetryManager.reset()` under method `stop` (or was it `close`).

src/logic/Node.ts
```diff
@@ -254,6 +261,8 @@ export class Node extends EventEmitter {
         const streamId = StreamIdAndPartition.fromMessage(instructionMessage)
         const { nodeIds, counter } = instructionMessage
 
+        this.instructionRetryManager.add(instructionMessage, trackerId)
```
harbu, line 264: Hmm.. a bit confused here now.

We set up `InstructionRetryManager` and pass it `handleTrackerInstruction` as callback function. OTOH here we invoke `instructionRetryManager.add` causing this function (`handleTrackerInstruction`) to be invoked again?

Also we have `InstructionThrottler` in the mix. I'm assuming `InstructionThrottler` should handle the message first and then pass it to `InstructionRetryManager`?

I'm wondering if we should organize the code so that the flow is more obvious:
```
onTrackerInstructionReceived(instruction) -> InstructionThrottler -> InstructionRetryManager -> handleTrackerInstruction
```

Would something like the following work (for the meantime):

In constructor:

```ts
this.instructionThrottler = new InstructionThrottler((instructionMessage, trackerId) => {
    this.instructionRetryManager.add(instructionMessage, trackerId)
})
this.instructionRetryManager = new InstructionRetryManager(this.handleTrackerInstruction.bind(this), 10000)
```

And then keep the following as is:
```ts
onTrackerInstructionReceived(trackerId: string, instructionMessage: TrackerLayer.InstructionMessage): void {
    this.instructionThrottler.add(instructionMessage, trackerId)
}
```

harbu (changes requested):
Now that I've gone thru the whole code, yeah I noticed what I'm missing. `InstructionRetryManager` is not used for the first handling of the instruction but rather the 2nd, 3rd, 4th, and so forth. I'd say, let's make the first call go thru the `InstructionRetryManager` as well to make it more consistent. My longer comment addresses this point.

harbu (approved):
Actually I think it works as is now. I'll create a ticket for the future for thinking about whether all instruction handling logic should be extracted into a class / function or smth.

harbu:
Added to backlog ticket about consolidating instruction handling: https://linear.app/streamr/issue/NET-210/refactor-instruction-handling-on-node-side, but it is not high priority imo.

PR #514, opened 2021-02-09
NET-194: fix latency metrics

harbu:
Latency metric was reported as a sum of latencies instead of last average latency. Fixed this by introducing a new metric type, "fixed" metric, in the metrics "framework". It is a metric that can be just simply set at any arbitrary point in time to any value and it will get reported as such.


src/helpers/MetricsContext.ts
```diff
@@ -63,6 +74,14 @@ export class Metrics {
...
     }
 
+    set(name: string, value: number): Metrics {
+        if (this.fixedMetrics[name] === undefined) {
+            throw new Error(`Not a fixed metric "${this.name}.${name}".`)
+        }
+        this.fixedMetrics[name] = value
+        return this
+    }
+
```
teogeb, line 84: LGTM

Would it make sense if all values would be sent to this class via the `record` method? In that case, this class would encapsulate the storage logic (store all vs. store the latest). The producer could just publish data to the `MetricsContext`, and it didn't need to care how the reporting module would consume it.

teogeb (approved):

src/helpers/MetricsContext.ts
harbu: Sort of. For the two modes, `FixedMetrics` and `RecordedMetrics` there really isn't a pressing need to have two separate methods to record values. For `QueriedMetrics` the story is a bit different since it is "pull-based".

PR #521, opened 2021-03-16
Per stream status

juslesan:
- No longer close connections on out of sync signalling messages
- Add logic for sending single stream statuses along with full node statuses when needed
- Sending of status messages has been significantly reduced.
- Tracker is using approximately 20-100x less cpu in local multi stream environment

A single node status is wrapped similarly to a full node status. Differentiation between full and stream statuses is done based on a boolean value. When a node sends a single stream status notifying the tracker of a unsubscription, it sets the status counter to -1.

### DIFF
--- a/src/logic/Node.ts
+++ b/src/logic/Node.ts
@@ -216,11 +216,10 @@ export class Node extends EventEmitter {
 
     unsubscribeFromStream(streamId: StreamIdAndPartition, sendStatus = true): void {
         this.logger.debug('unsubscribeFromStream: remove %s from streams', streamId)
-        const trackerId = this.getTrackerId(streamId)
         this.streams.removeStream(streamId)
         this.instructionThrottler.removeStreamId(streamId.key())
         this.instructionRetryManager.removeStreamId(streamId.key())
-        if (trackerId && sendStatus) {
+        if (sendStatus) {
             this.prepareAndSendStreamStatus(streamId)
         }
     }
--- a/src/logic/Node.ts
+++ b/src/logic/Node.ts
@@ -263,7 +263,7 @@ export class Node extends EventEmitter {
         const streamId = StreamIdAndPartition.fromMessage(instructionMessage)
         const { nodeIds, counter } = instructionMessage
 
-        this.instructionRetryManager.add(instructionMessage, trackerId)
+        // this.instructionRetryManager.add(instructionMessage, trackerId)
 
         // Check that tracker matches expected tracker
         const expectedTrackerId = this.getTrackerId(streamId)
@@ -423,7 +423,8 @@ export class Node extends EventEmitter {
             }),
             started: this.started,
             rtts: this.nodeToNode.getRtts(),
-            location: this.peerInfo.location
+            location: this.peerInfo.location,
+            singleStream: false
         }
     }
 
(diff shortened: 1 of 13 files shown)


src/logic/Node.ts
```diff
@@ -141,9 +141,11 @@ export class Node extends EventEmitter {
         this.nodeToNode.on(NodeToNodeEvent.DATA_RECEIVED, (broadcastMessage, nodeId) => this.onDataReceived(broadcastMessage.streamMessage, nodeId))
         this.nodeToNode.on(NodeToNodeEvent.NODE_DISCONNECTED, (nodeId) => this.onNodeDisconnected(nodeId))
         this.nodeToNode.on(NodeToNodeEvent.RESEND_REQUEST, (request, source) => this.requestResend(request, source))
-        this.on(Event.NODE_SUBSCRIBED, (nodeId, streamId) => {
+        this.on(Event.NODE_SUBSCRIBED, (nodeId, streamId, sendStatus = true) => {
```
harbu: Hmm.. we shouldn't really have to pass `sendStatus` here since it is a class internal matter. Would maybe work better if we remove this event handler and just do all this stuff directly in `subscribeToStreamOnNode` since it is the only place that emits this event.

src/logic/OverlayTopology.ts
```diff
@@ -86,6 +86,9 @@ export class OverlayTopology {
     }
 
     formInstructions(nodeId: string, forceGenerate = false): Instructions {
+        if (!this.nodes[nodeId]) {
```
harbu, line 89: Hmm.. did you run into this in practice or is this in here sort of "just in case"? If the latter, it may be worth logging a warning since it indicates a potential bug in our internal logic.

src/logic/Tracker.ts
```diff
@@ -173,19 +177,36 @@ export class Tracker extends EventEmitter {
...
+            const streamKey = Object.keys(streams)[0]
+            const status = streams[streamKey]
+            if (status.counter === -1) {
+                this.leaveAndCheckEmptyOverlay(streamKey, this.overlayPerStream[streamKey], node)
+            } else {
+                const neighbors = new Set([...status.inboundNodes, ...status.outboundNodes])
+                this.overlayPerStream[streamKey].update(node, [...neighbors])
+            }
+        } else {
+            this.logger.warn('Empty single stream Status received')
```
harbu, line 191: 👍

src/logic/Tracker.ts
```diff
@@ -173,19 +177,36 @@ export class Tracker extends EventEmitter {
...
-                this.metrics.record('instructionsSent', 1)
-                try {
-                    const counterValue = this.instructionCounter.setOrIncrement(nodeId, streamKey)
-                    await this.trackerServer.sendInstruction(nodeId, StreamIdAndPartition.fromKey(streamKey), newNeighbors, counterValue)
-                    this.logger.debug('sent instruction %j (%d) for stream %s to node %s', newNeighbors, counterValue, streamKey, nodeId)
-                } catch (e) {
-                    this.logger.error(`Failed to formAndSendInstructions to node ${nodeId}, streamKey ${streamKey}, because of ${e}`)
-                }
-            })
+            if (this.overlayPerStream[streamKey]) {
```
harbu, line 197: Hmm.. did you run into this in practice or is this in here sort of "just in case"? If the latter, it may be worth logging a warning since it indicates a potential bug in our internal logic.

harbu (approved):

PR #522, opened 2021-03-18
force connections on instruction reattempts

juslesan:
- During churn, nodes can end up with inconsistent stream and connection states. 
- Instruction reattempts have a high probability for being triggered at different times between nodes that are instructed to connect between each other. This causes connections to timeout
- This change forces connections between nodes by using RtcConnect messages during Instruction retries.
- Ensures  (at least more likely)  that nodes connect to each other based on instructions after the first instruction retry is ran

### DIFF
--- a/src/logic/Node.ts
+++ b/src/logic/Node.ts
@@ -255,7 +255,7 @@ export class Node extends EventEmitter {
         this.instructionThrottler.add(instructionMessage, trackerId)
     }
 
-    async handleTrackerInstruction(instructionMessage: TrackerLayer.InstructionMessage, trackerId: string): Promise<void> {
+    async handleTrackerInstruction(instructionMessage: TrackerLayer.InstructionMessage, trackerId: string, reattempt = false): Promise<void> {
         const streamId = StreamIdAndPartition.fromMessage(instructionMessage)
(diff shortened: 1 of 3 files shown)


src/logic/Node.ts
```diff
@@ -278,7 +278,11 @@ export class Node extends EventEmitter {
         const nodesToUnsubscribeFrom = currentNodes.filter((nodeId) => !nodeIds.includes(nodeId))
 
         const subscribePromises = nodeIds.map(async (nodeId) => {
-            await promiseTimeout(this.nodeConnectTimeout, this.nodeToNode.connectToNode(nodeId, trackerId))
+            if (reattempt) {
+                await promiseTimeout(this.nodeConnectTimeout, this.nodeToNode.connectToNode(nodeId, trackerId, true, false))
```
harbu, line 282: I guess a potential issue may be if both nodes end up taking the offerer role?

harbu (approved):
I'll need to test this one out as well

src/logic/Node.ts
juslesan: Yes, but it's much less likely than the previous issue with continuous timeouts. We should definetely discuss this problem with forming the webrtc connections on monday in the workshop

PR #524, opened 2021-03-23
rework forced webrtc connections

juslesan:
- Use RtcConnect message based on a new force connect boolean
- Ensures that all node-to-node connections are eventually formed based on the last tracker instruction


src/logic/Tracker.ts
```diff
@@ -184,7 +184,7 @@ export class Tracker extends EventEmitter {
                 this.overlayPerStream[streamKey].update(node, [...neighbors])
             }
         } else {
-            this.logger.warn('unexpected empty single-stream status received from node %s, contents %j', node, streams)
+            this.logger.debug('unexpected empty single-stream status received from node %s, contents %j', node, streams)
```
harbu, line 187: I'm seeing a lot of these btw when running integration tests

harbu (approved):

src/logic/Tracker.ts
juslesan: Yeah, this is triggered whenever a single stream status is filtered based on instruction counters

PR #525, opened 2021-03-24
try alternative ping/pong logic (cherry-picked)

harbu:
- Instances of `Connection` handle ping/pong timers independently (before it was managed by `WebRTCEndpoint`)
    - No need for `WebRTCEndpoint` to know anything about ping/pong details.
    - It was a bit weird that there was a retry-timer being managed by `Connection`  itself, but also `WebRTCEndpoint`. Possibly leading to pings going out a bit more often than intended?
 - Only retry pings (not pongs)  and disconnect if `pingAttempts >= maxPingPongAttempts`.
    - Before we would re-try pongs, now we only send a pong once and it either goes thru or it doesn't.  We don't use pong failures as a reason to disconnect, only ping failures. This can be done since we know that each end of the connection is sending pings of their own.

### DIFF
--- a/src/composition.ts
+++ b/src/composition.ts
@@ -174,8 +174,7 @@ function startNode({
             peerInfo,
             stunUrls,
             webRtcSignaller, 
-            metricsContext, 
-            pingInterval, 
+            metricsContext,
             newWebrtcConnectionTimeout,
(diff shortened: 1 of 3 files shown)


juslesan (approved):
LGTM

PR #526, opened 2021-03-25
Fix bin

juslesan:


harbu (approved):
Ahh forgot about all the /bin/ stuff again. :grimacing:  Thx for the fixes!

PR #527, opened 2021-03-25
 NET-141: Stream ids in Network API should be case-insensitive

bearni95:


harbu:
Closing this due to feature cancelled.

PR #540, opened 2021-04-12
Test types in CI

timoxley:
* Tests types in CI using new `npm run test-types` script.
* Allows running `npm run test-types -- --watch` locally to get simple type checking during development without running an IDE.
* Also enables incremental compilation, improves recompilation speed.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -34,6 +34,7 @@ jobs:
             ${{ runner.OS }}-
       - run: npm ci
       - run: npm run eslint
+      - run: npm run test-types
       - run: npm run build --if-present
       - run: npm test
         env:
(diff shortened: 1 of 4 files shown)


.github/workflows/nodejs.yml
```diff
@@ -34,6 +34,7 @@ jobs:
             ${{ runner.OS }}-
       - run: npm ci
       - run: npm run eslint
+      - run: npm run test-types
       - run: npm run build --if-present
```
harbu, line 38: Doesn't building the project `npm run build --if-present` already validate the types by compilation?

timoxley: yeah it does, could remove this I guess but with --incremental enabled now it shouldn't make much difference

harbu: Ah I see, make sense.

harbu (approved):

juslesan (approved):

PR #544, opened 2021-04-19
Prevent Connection reuse

timoxley:
Prevents reuse of a `Connection` object.

After `Connection` is closed/errored, a new `Connection` object must be created.
No events or callbacks should function after connection is closed.
This should help reduce possibility of async contamination between connections.

* Lifts `messageQueue` out of `Connection` and into `WebRtcEndpoint`.
* Closes connection on connection failed. 
* Logs previous state when logging connection/gathering state transitions


src/connection/Connection.ts
```diff
@@ -75,9 +77,10 @@ export class Connection {
         routerId,
         isOffering,
         stunUrls,
+        messageQueue,
         bufferThresholdHigh = 2 ** 17,
         bufferThresholdLow = 2 ** 15,
-        newConnectionTimeout = 8000,
+        newConnectionTimeout = 15000,
```
harbu, line 147: is the timeout still within reasonable limits, thoughts @juslesan ?

src/connection/Connection.ts
```diff
@@ -134,30 +138,44 @@ export class Connection {
     }
 
     connect(): void {
+        this.logger.info('connect()')
```
harbu: Do we want this at `info` level? Seems like it could lead to a lot of logs in production.

src/connection/WebRtcEndpoint.ts
```diff
@@ -262,6 +273,9 @@ export class WebRtcEndpoint extends EventEmitter {
         this.stopped = true
         Object.values(this.connections).forEach((connection) => connection.close())
         this.connections = {}
+
+        Object.values(this.messageQueues).forEach((queue) => queue.clear())
```
harbu: Only doing `this.messageQueues = {}` is not enough here?

test/integration/Connection.test.ts
```diff
@@ -172,7 +178,7 @@ describe('Connection', () => {
         expect(connectionTwo.getRtt()).toBeGreaterThanOrEqual(0)
     })
 
-    it('messages delivered again on temporary loss of connectivity', async () => {
+    it.skip('messages delivered again on temporary loss of connectivity', async () => {
```
harbu: skipped?

src/connection/Connection.ts
juslesan: In the AWS experiment branch I set this timeout up to 18000, increasing the timeout does come with challenges for fast (uninstructed) reconnects for sure though. I am not sure if the timeout needs to be this high though, the instruction handling timeout seems to be more important

juslesan: https://github.com/streamr-dev/network/pull/546#discussion_r616464166

test/integration/Connection.test.ts
juslesan: This test seems to be broken, this is likely due to the messageQueue being moved to webrtcEndpoint and thus the  test should likely be moved there

timoxley: Correct, yeah there's no "temporary loss of connectivity" for the connection now, since it's one-use only. I will try move it into the webrtcEndpoint

src/connection/WebRtcEndpoint.ts
timoxley: Perhaps, though we want to make sure:

1. messages are definitely wiped
2. anything that happens to have acquired a reference to the message queue can't continue to operate with it e.g. can't keep iterating.

Memory leaks aren't immediately obvious or easy to detect so doesn't hurt to try be very explicit about trying to free memory when

1. there potentially a lot of memory locked up (e.g. holding any user supplied data)
2. it's a one liner to at least remove the primary references to it

src/connection/Connection.ts
timoxley: Good point, have lowered this to debug in 2a5335dc

test/integration/Connection.test.ts
timoxley: Ahh @juslesan has already moved it in `various-fixes`, excellent 👍 

https://github.com/streamr-dev/network/pull/546/commits/c711b3dd18e2a2b043ad60619676c70619a4ee1d

src/connection/WebRtcEndpoint.ts
harbu: Good points, thanks for clarifying.

harbu (approved):

src/connection/Connection.ts
```diff
@@ -23,17 +24,60 @@ export interface ConstructorOptions {
...
+type HandlerParameters<T extends (...args: any[]) => any> = Parameters<Parameters<T>[0]>
+
+/**
+ * Create an EventEmitter that fires appropriate events for
+ * each peerConnection.onEvent handler.
+ *
+ * Wrapping allows us to trivially clear all event handlers.
+ * There's no way to reliably stop PeerConnection from running an event handler
+ * after you've passed it. Closing a connection doesn't prevent handlers from firing.
+ * Replacing handlers with noops doesn't work reliably, it can still fire the old handlers.
```
harbu, line 52: Kudos on explanation :+1:

src/connection/Connection.ts
```diff
@@ -57,6 +101,8 @@ export class Connection {
 
     private connection: PeerConnection | null
     private dataChannel: DataChannel | null
+    private dataChannelEmitter?: ReturnType<typeof DataChannelEmitter>
```
harbu: Does
```
ReturnType<typeof DataChannelEmitter>
```
just resolve to `EventEmitter` or does it validate that the value being assigned has to be indeed the return value of function `DataChannelEmitter`? If the previous, would it just be simpler to use `EventEmitter` as the type here?

src/connection/Connection.ts
```diff
@@ -123,56 +171,87 @@ export class Connection {
...
+
+        this.onStateChange = this.onStateChange.bind(this)
+        this.onLocalCandidate = this.onLocalCandidate.bind(this)
+        this.onLocalDescription = this.onLocalDescription.bind(this)
+        this.onStateChange = this.onStateChange.bind(this)
+        this.onGatheringStateChange = this.onGatheringStateChange.bind(this)
+        this.onDataChannel = this.onDataChannel.bind(this)
+    }
+
+    onStateChange(state: string): void {
```
harbu: Should these `on..` methods be made `private` in order to signal to the reader of the class that it is not intended that these methods be called by the user directly?

src/connection/Connection.ts
```diff
@@ -23,17 +24,60 @@ export interface ConstructorOptions {
...
+    connection.onGatheringStateChange((...args: HandlerParameters<PeerConnection['onGatheringStateChange']>) => (
+        emitter.emit('gatheringStateChange', ...args)
+    ))
+    connection.onLocalDescription((...args: HandlerParameters<PeerConnection['onLocalDescription']>) => emitter.emit('localDescription', ...args))
+    connection.onLocalCandidate((...args: HandlerParameters<PeerConnection['onLocalCandidate']>) => emitter.emit('localCandidate', ...args))
+    connection.onDataChannel((...args: HandlerParameters<PeerConnection['onDataChannel']>) => emitter.emit('dataChannel', ...args))
+    return emitter
+}
+
+function DataChannelEmitter(dataChannel: DataChannel) {
```
harbu, line 75: Sort of an enhancement idea but the return values of these could be typed so that simple spelling mistakes are caught by the compiler, e.g.:

```ts
const dce = DataChannelEmitter(this.dataChannel)
dce.on('opne', ....) // <- compilation error!
```

test/integration/Connection.test.ts
harbu: Can this test be removed now then?

src/connection/Connection.ts
harbu: Example of this here: https://github.com/streamr-dev/network/blob/master/src/logic/Node.ts#L54-L64

test/integration/Connection.test.ts
juslesan: It will be removed from this file once the various fixes PR is merged to this branch

juslesan (approved):

src/connection/Connection.ts
timoxley: Ahh you know what, I think this system may be simpler than `[strict-event-emitter-types](https://github.com/bterlson/strict-event-emitter-types)`, I wonder if there's a difference.

I'm going to have a look at the differences and probably remove `strict-event-emitter-types` to replace it with this approach.

src/connection/Connection.ts
timoxley: yeah it resolves to `EventEmitter`, I think you're right, it'd be simpler

src/connection/Connection.ts
timoxley: @harbu  Ahh, of course, `strict-event-emitter-types` also deals with all the other event emitter methods `.on`, `.once`, `removeListener`, etc

timoxley: https://github.com/bterlson/strict-event-emitter-types/blob/b653d2caba1970c206829a8cbaf679391328149a/src/index.ts

timoxley: So it has some overhead in the form of the `import` statement and the weird `EventEmiter as ({ new(): StrictEventEmitter<EventEmitter, Events> })` thing but gives more safety.

harbu: Given that it provides more safety, worth using then (even though one line of syntax weirdness)?

PR #546, opened 2021-04-19
Various Fixes

timoxley:
Found some suspicious looking code while hunting for source of connection bug.

* Fix some flakey async behaviour in `storage-config.test`. Seemed the event possibly fired before the listener was attached.
* ~~Actually compare `nodeIds` & `subscribeNodeIds` when determining whether tracker instruction fulfilled. Currently it just checks the number of nodes are the same, rather than the actual node ids.~~ Not needed.
* Nullify connection event handlers if connection is no longer current. Ensures old connections can't interfere with new state.
* Resolve subsequent calls to `connect` with a promise that resolves the same way the initial call does.
* Increase default `newConnectionTimeout` from 5000 to 15000. @juslesan mentioned connections can sometimes take a long time to form.
* Delete connection immediately when instructing it to close.
* Early return in propagateMessage when no subscribers to avoid dangling else. 
* close/clear endpoint connections/queues after resetting everything else.


src/logic/Node.ts
```diff
@@ -111,7 +111,7 @@ export class Node extends EventEmitter {
         this.bufferTimeoutInMs = opts.bufferTimeoutInMs || 60 * 1000
         this.bufferMaxSize = opts.bufferMaxSize || 10000
         this.disconnectionWaitTime = opts.disconnectionWaitTime || 30 * 1000
-        this.nodeConnectTimeout = opts.nodeConnectTimeout || 4000
+        this.nodeConnectTimeout = opts.nodeConnectTimeout || 15000
```
juslesan, line 114: This is likely a necessary increase based on the AWS experiments.
https://github.com/streamr-dev/network/pull/544#discussion_r616458096

src/connection/Connection.ts
```diff
@@ -60,7 +60,7 @@ export class Connection {
     private connection: PeerConnection | null
     private dataChannel: DataChannel | null
     private paused: boolean
-    private lastState: string | null
+    public lastState: string | null
```
harbu: Keeping this private and providing a getter to fetch its value provides more guarantees about the value being only modified within this class, which helps the reader of this class reason about where and when this value is modified.

test/integration/storage-config.test.ts
```diff
@@ -64,13 +64,17 @@ describe('storage node', () => {
             storages: [storage as Storage],
             storageConfig: config
         })
+
+        // @ts-expect-error private field
+        const t1 = waitForEvent(tracker.trackerServer, TrackerServerEvent.NODE_STATUS_RECEIVED)
```
harbu, line 69: Did this alleviate the flakyness then?

src/connection/WebRtcEndpoint.ts
```diff
@@ -147,7 +147,7 @@ export class WebRtcEndpoint extends EventEmitter {
             })
     }
 
-    connect(
+    async connect(
```
harbu, line 154: why `async`, `await` not used?

src/connection/WebRtcEndpoint.ts
```diff
@@ -175,17 +187,29 @@ export class WebRtcEndpoint extends EventEmitter {
             newConnectionTimeout: this.newConnectionTimeout,
             pingInterval: this.pingInterval,
             onLocalDescription: (type, description) => {
+                if (this.connections[targetPeerId] !== connection) {
```
harbu: What is the relationship with the guards here and the ones you wrote here https://github.com/streamr-dev/network/pull/544/files#diff-4e51e534b6625d3469234524fa87f3e6c107ffe5c460926d93b001e939d3561dR140-R180. Are all these guards needed?

src/logic/Node.ts
```diff
@@ -320,9 +320,12 @@ export class Node extends EventEmitter {
         this.logger.debug('subscribed to %j and unsubscribed from %j (streamId=%s, counter=%d)',
             subscribeNodeIds, unsubscribeNodeIds, streamId, counter)
 
-        if (subscribeNodeIds.length !== nodeIds.length) {
+        if (JSON.stringify(subscribeNodeIds.slice().sort()) !== JSON.stringify(nodeIds.slice().sort())) {
```
harbu: Why would the node ids be different though? The values of `subscribeNodeIds`  will always be a subset of `nodeIds` (unless there is a bug elsewhere in the code)?

src/logic/Node.ts
```diff
@@ -383,52 +386,57 @@ export class Node extends EventEmitter {
...
+        if (!subscribers.length) {
+            this.logger.warn('put back to buffer because could not propagate to %d nodes or more: %j',
+                MIN_NUM_OF_OUTBOUND_NODES_FOR_PROPAGATION, streamMessage.messageId)
+            this.logger.debug('streams: %j', this.streams)
             this.seenButNotPropagatedSet.add(streamMessage)
             this.messageBuffer.put(streamIdAndPartition.key(), [streamMessage, source])
+            return
         }
+
+        this.logger.info('propagating to subscribers: %s', subscribers.join(', '))
```
harbu: Again, wondering if `info` should be here the default w.r.t. production logging needs

src/logic/Node.ts
```diff
@@ -383,52 +386,57 @@ export class Node extends EventEmitter {
...
+        subscribers.forEach(async (subscriber) => {
+            try {
+                this.logger.info('propagating to %s >', subscriber)
```
harbu: same on logging levels

src/connection/Connection.ts
timoxley: This seems awfully Java but fair point, if it's private it can't possibly be mutated externally.

test/integration/storage-config.test.ts
timoxley: Yes, it appeared to, though maybe I just got lucky 3x in a row? There's actually a bunch of similar potential issues in the tests, one solution would be to wait for event loop to turn before starting anything that could trigger an event so callers always have a chance to attach listeners.

src/connection/WebRtcEndpoint.ts
timoxley: Any time you're returning a promise you should always use `async`, since this effectively wraps the whole function in a `try { … } catch(err) { return Promise.reject(err) }`. 

Otherwise you can end up with accidental uncaught sync errors e.g.

```js
function setup() {
  throw new Error('oops') // unintentional error
}

function connect() {
  setup()
  return new Promise((resolve) => resolve())
}

connect().catch(console.log) // won't catch because setup() threw synchronously
```

Should we crash for unintentional sync exceptions? Perhaps, but the issue is that it's very easy to accidentally create sync errors e.g. add some argument validation in `setup()` or something.

For the least surprises, when a function returns promises i.e. is async, it should only ever return promises, errors should always become rejections i.e. [Do not release Zalgo](https://blog.izs.me/2013/08/designing-apis-for-asynchrony). 

Adding `async` is low cost, but adds a *guarantee* that the function will not throw synchronously, ever, and the writer nor consumer will never have to think about whether that case is even possible.

src/connection/WebRtcEndpoint.ts
timoxley: Yeah perhaps not all needed, though I didn't see anything that made it obvious that it was 100% impossible for either:

1. a connection to be finished and not (yet) removed from list of connections
2. a connection to be removed from list of connections and not (yet) finished

I'll take another look into cleaning these up without need to keep checking the guards in all cases, agreed it's error prone and a smell.

src/logic/Node.ts
timoxley: ahh yes I see, you're right. Maybe just needs a comment, looked like a weird assumption but now I see it makes sense.

src/logic/Node.ts
timoxley: Will reduce to `debug`

src/logic/Node.ts
timoxley: Removed these logs entirely.

src/connection/WebRtcEndpoint.ts
timoxley: Converted `Connection` into an `EventEmitter` in 608fbc71

Same deal as the previous change to wrap `PeerConnection`/`DataChannel` event handling in an `EventEmitter`, makes it trivial to turn off all listeners immediately with `removeAllListeners`.

src/connection/Connection.ts
timoxley: Done.

timoxley:
Updated, needs re-review, in particular note the connection timeout gets reset when state moves to `connecting`, issue being I was seeing multiple second delays between connection state `null` (creation) and `connecting`:
https://github.com/streamr-dev/network/pull/546/commits/5c1afa45642651bfed213e8ec4179d0cc8c3a292

Not sure why, there was no logging from that side during the delay but might be that it needs to wait for other end to time out before it can try again, so it always needs 2x timeout? Not sure.

src/connection/WebRtcEndpoint.ts
harbu: I haven't thought about it from this angle before, but it makes sense to me now. Essentially the function's interface becomes easier to work with, since instead of having to

- (a) expect a sync error to be thrown
- (b) expect a rejection error
- (c) expect resolved value

the first case goes away, and you are left as the user of the function to handle only (b) and (c). Did I understand correctly @timoxley ?

src/connection/Connection.ts
```diff
@@ -73,7 +66,26 @@ function DataChannelEmitter(dataChannel: DataChannel) {
...
+    message: (msg: string)  => void
+    close: (err?: Error) => void
+    error: (err: Error) => void
+    bufferLow: () => void
+    bufferHigh: () => void
+}
+
+// reminder: only use Connection emitter for external handlers
+// to make it safe for consumers to call removeAllListeners
+// i.e. no this.on('event')
```
harbu, line 102: :+1:

src/connection/Connection.ts
```diff
@@ -73,7 +66,26 @@ function DataChannelEmitter(dataChannel: DataChannel) {
...
+export const ConnectionEmitter = EventEmitter as { new(): StrictEventEmitter<EventEmitter, Events> }
```
harbu, line 103: Having trouble wrapping my head around this type definition.

Am I understanding this correctly: you are using `EventEmitter`  as the base type but then extending it so that its constructor will result in objects of type `StrictEventEmitter<EventEmitter, Events>`?

src/connection/Connection.ts
```diff
@@ -428,6 +385,45 @@ export class Connection {
         }
     }
 
+    private onStateChange(state: string): void {
```
harbu, line 403: Ah looks like this and the other PR has some duplication. Looks like the methods here are private which I was asking about on the other PR.

src/connection/WebRtcEndpoint.ts
```diff
@@ -99,7 +99,7 @@ export class WebRtcEndpoint extends EventEmitter {
                 connection.setPeerInfo(PeerInfo.fromObject(originatorInfo))
                 connection.setRemoteDescription(description, 'answer' as DescriptionType.Answer)
             } else {
-                this.logger.warn('unexpected rtcAnswer from %s: %s', originatorInfo, description)
+                this.logger.warn('unexpected rtcAnswer from %s: %s', peerId, description)
```
harbu, line 102: I have thought it better to move towards printing `peerInfo` details in logs instead of just ids because the ids are Ethereum addresses which makes it hard to follow what is going on. PeerInfo#toString will print the peer name as well as the first characters of the address: https://github.com/streamr-dev/network/blob/af6d1c571ce27f5b3ce741e4876d5a45df76f6c0/src/connection/PeerInfo.ts#L82-L84

harbu: In the tests of network, ids are set artificially to be any string but in the broker-end they are always Ethereum addresses.

src/connection/WebRtcEndpoint.ts
```diff
@@ -109,21 +109,25 @@ export class WebRtcEndpoint extends EventEmitter {
...
             this.connect(peerId, routerId, isOffering).catch((err) => {
-                this.logger.warn('connectListener induced connection failed, reason %s', err)
+                this.logger.warn('connectListener induced connection from %s failed, reason %s', peerId, err)
             })
         })
 
         rtcSignaller.setErrorListener(({ targetNode, errorCode }: ErrorOptions) => {
             const error = new WebRtcError(`RTC error ${errorCode} while attempting to signal with ${targetNode}`)
-            this.emit(`errored:${targetNode}`, error)
+            const connection = this.connections[targetNode]
```
harbu, line 126: Hmm.. this change has some consequences, but I'm guessing it will be more on the side of fixing issues.

harbu (approved):

juslesan (approved):

src/connection/Connection.ts
timoxley: Sort of, more that it's `extending EventEmitter` but overriding (not extending) the type so it defines a constructor (`new`) that gives you back an `EventEmitter` extended by `StrictEventEmitter`.

https://github.com/bterlson/strict-event-emitter-types#usage-with-subclasses

src/connection/WebRtcEndpoint.ts
timoxley: That's a good idea.

src/connection/Connection.ts
timoxley: Related: https://github.com/streamr-dev/network/pull/544#discussion_r619080467

src/connection/WebRtcEndpoint.ts
timoxley: Ideally we'd use `peerInfo` everywhere instead of strings, but currently many places only have one of the name/id or the address, but not both. 
I tried changing `Connection` to take `PeerInfo` objects instead of strings for `selfId` and `targetPeerId`, and was able to make this change work pretty straightforwardly.

But then things got tricky once I was in `handleTrackerInstruction` calling `nodeToNode.connectToNode` in `logic/Node.ts`, as it wasn't immediately clear how to get the bits needed to build a `PeerInfo`, so I just dropped it.

The logging changes here are mostly about fixing `from: [object Object]` from appearing in the logs, let's land this as-is.

src/connection/WebRtcEndpoint.ts
timoxley: 100% correct 👍

PR #547, opened 2021-04-23
NET-264: remove node-level buffering

harbu:
Remove buffering from `Node` (logic layer). 

The following changes to preexisting public API methods have been made.
- Method `publish(msg: StreamMessage): number` is now the designated "fire-and-forget" publish mechanism. It returns the number of neighbors the message was forwarded (attempted) to, which could server as a useful signal in certain cases.

The following new methods have been added to the public API.
- Method `asyncPublish(msg: StreamMessage, minNeighbors = 1, ...): Promise<number>` is a more "reliable" publish mechanism. It will wait for at least neighbors to be present before attempting to propagate. It too, signals the number of neighbors the message was forwaded (attempted) to. It will reject, if the minimum number of neighbors were not available within a given timeout.
- Method `getNumberOfNeighbors(streamId: string, streamPartition: number): number` returns the number of available neighbors for a stream at the present time.
- Method `waitForNeighbors(streamId: string, streamPartition: number, minNeighbors = 1, ...): Promise<number>` similar thinking to `asyncPublish` but we do not publish but just wait.

### Caveats
- Buffering will most likely need to be implemented on broker in order to buffer incoming messages from clients.
- Parameter `minNeighbors` is quite arbitrary and there is not a "correct" answer to what it should be currently (There are valid cases where there are no neighbors, however, setting it as zero means neighbors are not awaited for :shrug:.) The ideal would be not to have to supply such a parameter, however, that would require some significant changes on how the tracker sends instructions which I am hesitant to do right now.


juslesan (approved):
LGTM!

harbu:
Moved this PR to network-monorepo, see: https://github.com/streamr-dev/network-monorepo/pull/16

PR #548, opened 2021-04-27
NET-129: fix flaky test

juslesan:
1. Update node-datachannel to v0.1.2
2. Check that datachannel is still open immediately after a send attempt. If isOpen() returns false and sendMessage() returns true messages were lost at a ~100% rate. It is better to to expect that messages are lost and to attempt resends after such cases. This approach fixes the flaky test case "test/unit/WebRtcEndpoint.test.ts:  messages are delivered on temporary loss of connectivity"
3. Wrapping the node-datachannel.sendMessage() call with an if check for isOpen() does not fix the issue as it is likely timing related

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -517,8 +517,9 @@ export class Connection extends ConnectionEmitter {
             } else {
                 let sent = false
                 try {
-                    // Checking `this.open()` is left out on purpose. We want the message to be discarded if it was not
-                    // sent after MAX_TRIES regardless of the reason.
+                    // this.isOpen() is checked immediately after the call to node-datachannel.sendMessage() as if
+                    // this.isOpen() returns false after a "successful" send, the message is lost with a near 100% chance.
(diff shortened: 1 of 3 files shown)


harbu (approved):

src/connection/Connection.ts
```diff
@@ -529,6 +529,7 @@ export class Connection extends ConnectionEmitter {
                 if (sent) {
                     this.messageQueue.pop()
                     queueItem.delivered()
+                    numOfSuccessSends += 1
                 } else {
                     this.processFailedMessage(queueItem, new Error('sendMessage returned false'))
```
timoxley, line 534: @juslesan I wonder if this message might now be misleading

juslesan: Sure it might be. Although, once the attemptToFlushMessages() function is called again after the error, the next error should be due to the library returning false. (Assuming that the next call gets this far)

PR #553, opened 2021-05-04
update node-datachannel, make WebRTC endpoint unit tests more stable

juslesan:
- General stability updates for the WebRTC endpoint
- WebRTC endpoint tests fast reconnect tests much less flaky on local Ubuntu 20 VM (More instability in CI)

### DIFF
--- a/src/connection/Connection.ts
+++ b/src/connection/Connection.ts
@@ -4,7 +4,6 @@ import nodeDataChannel, { DataChannel, DescriptionType, LogLevel, PeerConnection
 import { Logger } from '../helpers/Logger'
 import { PeerInfo } from './PeerInfo'
 import { MessageQueue, QueueItem } from './MessageQueue'
-import { Node } from "../logic/Node"
 
 nodeDataChannel.initLogger("Error" as LogLevel)
 
@@ -131,7 +130,7 @@ export class Connection extends ConnectionEmitter {
     private flushTimeoutRef: NodeJS.Timeout | null
     private connectionTimeoutRef: NodeJS.Timeout | null
     private pingTimeoutRef: NodeJS.Timeout | null
-    private flushRef: NodeJS.Timeout | NodeJS.Immediate | null
+    private flushRef: NodeJS.Immediate | null
     private pingAttempts = 0
     private rtt: number | null
     private rttStart: number | null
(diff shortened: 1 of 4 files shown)


src/connection/Connection.ts
```diff
@@ -443,12 +446,11 @@ export class Connection extends ConnectionEmitter {
         this.paused = false
         this.dataChannelEmitter = DataChannelEmitter(dataChannel)
         dataChannel.setBufferedAmountLowThreshold(this.bufferThresholdLow)
-        if (this.isOffering) {
-            this.dataChannelEmitter.on('open', () => {
-                this.logger.debug('dc.onOpen')
-                this.openDataChannel(dataChannel)
-            })
-        }
+        this.dataChannelEmitter.on('open', () => {
```
harbu, line 449: Is this if the other side ends up offering after all?

harbu (approved):

src/connection/Connection.ts
juslesan: This is a reaction to a change in the libdatachannel API. In the WebRTC standard both ends should be able to call onOpen regardless of who is offering. This was not possible previously.

src/connection/Connection.ts
```diff
@@ -301,6 +301,9 @@ export class Connection extends ConnectionEmitter {
             }
         }
 
+        if (this.flushRef) {
+            clearImmediate(this.flushRef)
+        }
```
timoxley, line 306: Good catch

timoxley (approved):
Nice one.

@juslesan do you know or have any theory as to exactly why this improves stability

juslesan:
> Nice one.
> 
> @juslesan do you know or have any theory as to exactly why this improves stability

@timoxley 

Obviously there's been a punch of updates in libdatachannel between 0.1.2 and 0.1.3 of node-datachannel. It appeared that the never version was stabler in local VM. Also, the onDatachannel() event handler being available for both ends seems to have been an improvement. After using the handler on both ends the tests passed at much higher rates on local VM. Atleast the fast reconnect test case seems to be much less flaky than before.

Currently I'm 99% certain that the flakiness of test case 'messages are delivered on temporary loss of connectivity' is timing related and the issue is likely caused by the underlying libraries. After adding a console.log() for debugging before each send call the test was no longer flaky. Maybe there needs to be a timeout before flushing messages on a new datachannel. However, my most recent attempt at it did not fix the problem in CI.

PR #556, opened 2021-05-07
Net-25: node-to-node connections communicate and use the latest version supported by both

juslesan:
- Node-to-Node (WebRTC) connections exchange information of their control and message layer protocol versions
- Latest common protocol version is used to package messages in node-to-node connections
- Nodes disconnect if a shared protocol version is not found
- Tracker now maintains the used protocol versions of all connected nodes
- Requires this update to signalling messages in protocol  https://github.com/streamr-dev/streamr-client-protocol-js/pull/61


src/connection/NegotiatedProtocolVersions.ts
```diff
@@ -0,0 +1,57 @@
+import { PeerInfo } from "./PeerInfo"
+import { ControlLayer, MessageLayer } from "streamr-client-protocol"
+
+const defaultControlLayerVersions = ControlLayer.ControlMessage.getSupportedVersions()
+const defaultMessageLayerVersions = MessageLayer.StreamMessage.getSupportedVersions()
+
+type negotiatedProtocolVersion = { controlLayerVersion: number, messageLayerVersion: number }
```
timoxley: aren't Types conventionally CamelCase (uppercase first letter), unless they're primitives like `string`? Doesn't really matter just thought this looked odd

src/connection/NegotiatedProtocolVersions.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export class NegotiatedProtocolVersions {
+
+    private readonly peerInfo: PeerInfo
+    private readonly negotiatedProtocolVersions: { [key: string]: negotiatedProtocolVersion }
+
+    constructor(peerInfo: PeerInfo) {
+        this.negotiatedProtocolVersions = {}
```
timoxley: `Object.create(null)`

src/connection/NegotiatedProtocolVersions.ts
```diff
@@ -0,0 +1,57 @@
...
+            controlLayerVersion,
+            messageLayerVersion
+        }
+    }
+
+    removeNegotiatedProtocolVersion(peerId: string): void {
+        delete this.negotiatedProtocolVersions[peerId]
+    }
+
+    getNegotiatedProtocolVersion(peerId: string): negotiatedProtocolVersion | null | never {
```
timoxley: curious why `| never`

src/connection/NegotiatedProtocolVersions.ts
```diff
@@ -0,0 +1,57 @@
...
+        return this.negotiatedProtocolVersions[peerId] || null
+    }
+
+    validateProtocolVersions(controlLayerVersions: number[], messageLayerVersions: number[]): number[] {
+        if (controlLayerVersions === undefined || messageLayerVersions === undefined || controlLayerVersions === [] || messageLayerVersions === []) {
+            throw new Error('Missing version negotiation! Must give controlLayerVersions and messageLayerVersions as query parameters!')
```
timoxley, line 49: `arr === []` will never be true because `[]` generates a new object, and `===` checks object equality.

You probably want to check array length:

`(!controlLayerVersions || controlLayerVersions.length === 0) && (! messageLayerVersions || messageLayerVersions.length === 0)`

src/connection/PeerInfo.ts
```diff
@@ -55,9 +72,17 @@ export class PeerInfo {
         if (!Object.values(PeerType).includes(peerType)) {
             throw new Error(`peerType ${peerType} not in peerTypes list`)
         }
+        if (!controlLayerVersions || controlLayerVersions === []) {
```
timoxley: Same here, can't do `=== []`

src/connection/WebRtcEndpoint.ts
```diff
@@ -201,7 +219,7 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
                 // connection.
                 delete this.connections[targetPeerId]
             }
-
+            this.negotiatedProtocolVersions.removeNegotiatedProtocolVersion(targetPeerId)
```
timoxley, line 210: hm… is it possible that this will wipe out NegotiatedProtocolVersion information for another connection to the same peer? Maybe it should be inside the `if (this.connections[targetPeerId] === connection) {` above?

src/connection/WsEndpoint.ts
```diff
@@ -480,8 +495,18 @@ export class WsEndpoint extends EventEmitter implements IWsEndpoint {
         return this.peerBook.getAddress(peerId)
     }
 
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    getNegotiatedMessageLayerProtocolVersionOnNode(peerId: string): number | null {
```
timoxley: `: number | null`

I don't think `Math.max` can return `null`, but it can return `-Infinity` (`Math.max()`) or `NaN` (`Math.max('a')`). Also note `Math.max(null)` is `0`.

src/connection/WsEndpoint.ts
```diff
@@ -480,8 +495,18 @@ export class WsEndpoint extends EventEmitter implements IWsEndpoint {
...
+        return Math.max(...this.peerInfo.messageLayerVersions)
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    getNegotiatedControlLayerProtocolVersionOnNode(peerId: string): number | null {
+        return Math.max(...this.peerInfo.controlLayerVersions)
```
timoxley: hm, so why doesn't this need to use `peerId`?

src/protocol/NodeToNode.ts
```diff
@@ -71,7 +71,9 @@ export class NodeToNode extends EventEmitter {
     }
 
     send<T>(receiverNodeId: string, message: T & ControlLayer.ControlMessage): Promise<T> {
-        return this.endpoint.send(receiverNodeId, message.serialize()).then(() => message)
+        const controlLayerVersion = this.getNegotiatedControlLayerProtocolVersionOnNode(receiverNodeId)
+        const messageLayerVersion = this.getNegotiatedMessageLayerProtocolVersionOnNode(receiverNodeId)
```
timoxley: Wonder if maybe these could be a single function that returns a tuple with both versions, since most/all of the time you'll want both?

```js
const [controlLayerVersion, messageLayerVersion] = getNegotiatedProtocolVersionsOnNode(receiverNodeId)
```

src/protocol/NodeToNode.ts
```diff
@@ -127,4 +129,13 @@ export class NodeToNode extends EventEmitter {
     getRtts(): Readonly<Rtts> {
         return this.endpoint.getRtts()
     }
+
+    getNegotiatedMessageLayerProtocolVersionOnNode(peerId: string): number {
+        return this.endpoint.getNegotiatedMessageLayerProtocolVersionOnNode(peerId)
+            || Math.max(...this.endpoint.getPeerInfo().messageLayerVersions)
```
timoxley: curious about the scenario that needs the `||`

timoxley: Note if it's possible for `...this.endpoint.getPeerInfo().messageLayerVersions` to be `[]` then `Math.max()` will give `-Infinity`. Might not be possible, but since versions should always be >= 0, you can easily throw a `0` in there to prevent this from ever returning `-Infinity` i.e. `Math.max(0, ...this.endpoint.getPeerInfo().messageLayerVersions)`

src/connection/WsEndpoint.ts
```diff
@@ -369,9 +377,16 @@ export class WsEndpoint extends EventEmitter implements IWsEndpoint {
...
                     const peerId = res.headers['streamr-peer-id'] as string
                     const peerType = res.headers['streamr-peer-type'] as PeerType
+                    const controlLayerVersions = res.headers['control-layer-versions'] as string
+                    const messageLayerVersions = res.headers['message-layer-versions'] as string
 
-                    if (peerId && peerType) {
-                        serverPeerInfo = new PeerInfo(peerId, peerType)
+                    if (peerId && peerType && controlLayerVersions && messageLayerVersions) {
+                        const controlLayerVersionsArray = controlLayerVersions.split(',').map((version) => parseInt(version))
+                        const messageLayerVersionsArray = messageLayerVersions.split(',').map((version) => parseInt(version))
```
timoxley: Possible you want to filter out funky values like `NaN` and `-Infinity` here?

Also might want to add `10` as the radix i.e. `parseInt(version, 10)`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#octal_interpretations_with_no_radix

src/connection/PeerBook.ts
```diff
@@ -8,57 +8,36 @@ export class NotFoundInPeerBookError extends Error {
...
-        return this.addressToId[address]
-    }
-
-    hasAddress(address: string): boolean {
-        return this.addressToId[address] != null
-    }
-
-    hasPeerId(peerId: string): boolean {
-        return this.idToAddress[peerId] != null
+        return peerId.peerId
```
timoxley: Maybe the `peerId` variable here should be called `peerInfo` so then this reads like `peerInfo.peerId`

src/connection/PeerBook.ts
```diff
@@ -8,57 +8,36 @@ export class NotFoundInPeerBookError extends Error {
 }
 
 export class PeerBook {
```
timoxley, line 10: I like what you've done here, much tidier.

src/protocol/NodeToNode.ts
timoxley: Or even better, returns a `NegotiatedProtocolVersions`, but add the below magic to it so the above iterable destructuring still works:
```js
class NegotiatedProtocolVersions {
    // …
    [Symbol.iterator]() {
        return [this.controlLayerVersion, this.messageLayerVersion][Symbol.iterator]()
    }
}
```
or could make `controlLayerVersion` and `messageLayerVersion` into `public readonly` then use object destructuring:
```js
const { controlLayerVersion, messageLayerVersion } = getNegotiatedProtocolVersionsOnNode(receiverNodeId)
```

timoxley: or return a `NegotiatedProtocolVersions.negotiatedProtocolVersion` type object.

Or do nothing, just some ideas, your call, `getNegotiatedControlLayerProtocolVersionOnNode` and `getNegotiatedMessageLayerProtocolVersionOnNode` are quite a mouthful though

timoxley (changes requested):
The `=== []` checks need to be fixed but otherwise lgtm.

src/connection/NegotiatedProtocolVersions.ts
timoxley: https://linear.app/streamr/issue/NET-270/using-as-a-dictionary-creates-potential-for-security-issues-or-bugs

src/connection/NegotiatedProtocolVersions.ts
```diff
@@ -0,0 +1,57 @@
...
+    constructor(peerInfo: PeerInfo) {
+        this.negotiatedProtocolVersions = {}
+        this.peerInfo = peerInfo
+    }
+
+    addNegotiatedProtocolVersion(peerId: string, controlLayerVersion: number, messageLayerVersion: number): void {
+        this.negotiatedProtocolVersions[peerId] = {
+            controlLayerVersion,
+            messageLayerVersion
+        }
```
timoxley, line 32: Could this maybe be a `NegotiatedProtocolVersions` object

src/connection/NegotiatedProtocolVersions.ts
```diff
@@ -0,0 +1,57 @@
+import { PeerInfo } from "./PeerInfo"
+import { ControlLayer, MessageLayer } from "streamr-client-protocol"
+
+const defaultControlLayerVersions = ControlLayer.ControlMessage.getSupportedVersions()
+const defaultMessageLayerVersions = MessageLayer.StreamMessage.getSupportedVersions()
+
+type negotiatedProtocolVersion = { controlLayerVersion: number, messageLayerVersion: number }
+
+export class NegotiatedProtocolVersions {
```
timoxley, line 9: Wondering if maybe the `max` calculations etc could live in here.

test/integration/node-to-node-protocol-version-negotiation.test.ts
```diff
@@ -0,0 +1,136 @@
...
+        expect(ep1.getNegotiatedControlLayerProtocolVersionOnNode('node-endpoint2')).toEqual(null)
+        expect(ep2.getNegotiatedControlLayerProtocolVersionOnNode('node-endpoint1')).toEqual(null)
+    })
+
+    it('if there are no shared versions the connection is closed', async () => {
+        let errors = 0
+        try {
+            await ep3.connect('node-endpoint1', 'tracker')
+        } catch (err) {
+            errors += 1
```
timoxley, line 130: Doesn't need to change but just a headsup and I don't always follow this myself, but:

I recommend always checking the is the error you expect.

I've seen a number of cases where the error path had some bug that caused something like an "undefined is not a function" or something other than the expected error, and this went undetected because I didn't check the error type in the error handler. Maybe less of a concern with typescript but isn't expensive to set up the check:

```js
await expect(async () => (
    ep3.connect('node-endpoint1', 'tracker')
)).rejects.toThrow('no shared version') // matches against message, can be partial string match or even a regex
```

test/integration/node-to-node-protocol-version-negotiation.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    it('protocol versions are correctly negotiated',  () => {
+        expect(nodeToNode1.getNegotiatedControlLayerProtocolVersionOnNode('node-endpoint2')).toEqual(2)
+        expect(nodeToNode2.getNegotiatedControlLayerProtocolVersionOnNode('node-endpoint1')).toEqual(2)
+        expect(nodeToNode1.getNegotiatedMessageLayerProtocolVersionOnNode('node-endpoint2')).toEqual(31)
+        expect(nodeToNode2.getNegotiatedMessageLayerProtocolVersionOnNode('node-endpoint1')).toEqual(31)
+    })
+
+    it('messages are sent with the negotiated protocol version', (done) => {
+        ep2.on(wrtcEvent.MESSAGE_RECEIVED, (peerInfo, data) => {
```
timoxley: suggest using `ep2.once` so this can't possibly run more than once

.idea/jsLibraryMappings.xml
```diff
@@ -1,8 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
harbu: Intentional removal?

src/connection/NegotiatedProtocolVersions.ts
juslesan: Good catch, I just copied this from the old NET-25 PR and didn't give it much though

test/integration/node-to-node-protocol-version-negotiation.test.ts
juslesan: The problem here is that the error can change. Sometimes it's a connection timeout error when the next signalling message is never received and sometimes its a 'no shared version' error. Basically, it seems to mostly depend on timing.

It should be possible to check for different types of errors of course but in this case the most important point is that the connection is closed

timoxley: understood

timoxley (approved):
lgtm

src/connection/NegotiatedProtocolVersions.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    removeNegotiatedProtocolVersion(peerId: string): void {
+        delete this.negotiatedProtocolVersions[peerId]
+    }
+
+    getNegotiatedProtocolVersion(peerId: string): NegotiatedProtocolVersion | null {
+        return this.negotiatedProtocolVersions[peerId] || null
+    }
+
+    validateProtocolVersions(controlLayerVersions: number[], messageLayerVersions: number[]): number[] {
```
harbu: Looking at the usages of method `validateProtocolVersions` it always precedes an invocation to method `addNegotiatedProtocolVersion`. Thus I would turn this private, and make `addNegotiatedProtocolVersion` call this method before adding a protocol version to the object. This has the added benefit on ensuring that the following invariant is always true: any version present in field `this.negotiatedProtocolVersions` has undergone validation.

src/connection/PeerBook.ts
```diff
@@ -8,57 +8,36 @@ export class NotFoundInPeerBookError extends Error {
...
     add(peerAddress: string, peerInfo: PeerInfo): void {
-        const { peerId, peerType, peerName } = peerInfo
-        this.idToAddress[peerId] = peerAddress
-        this.addressToId[peerAddress] = peerId
-        this.addressToType[peerAddress] = peerType
-        this.addressToName[peerAddress] = peerName
+        this.peerInfos[peerAddress] = peerInfo
     }
 
     getPeerInfo(peerAddress: string): PeerInfo | null | never {
```
harbu, line 20: Hmm `never` can be removed from type signature. Maybe this used to throw an error or something?

src/connection/WebRtcEndpoint.ts
```diff
@@ -180,6 +191,13 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
         })
         connection.once('localDescription', (type, description) => {
             this.rtcSignaller.onLocalDescription(routerId, connection.getPeerId(), type, description)
+            try {
```
harbu: Duplicated code fragment with above, move to helper method in case future changes needed?

src/connection/WebRtcEndpoint.ts
```diff
@@ -269,6 +287,14 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
         return this.peerInfo
     }
 
+    getNegotiatedMessageLayerProtocolVersionOnNode(peerId: string): number | null {
+        return this.negotiatedProtocolVersions.getNegotiatedProtocolVersion(peerId)?.messageLayerVersion || null
```
harbu: Don't think the `|| null` is strictly needed here and below. Since the "elvis" operator  here `this.negotiatedProtocolVersions.getNegotiatedProtocolVersion(peerId)?.` will return a null if `this.negotiatedProtocolVersions.getNegotiatedProtocolVersion(peerId)` is a null. And if we get a `messageLayerVersion` it will always be a number based on its type.

src/connection/PeerInfo.ts
```diff
@@ -10,39 +11,55 @@ export enum PeerType {
...
     location?: Location | null | undefined
 }
 
+const defaultControlLayerVersions = ControlLayer.ControlMessage.getSupportedVersions()
+const defaultMessageLayerVersions = MessageLayer.StreamMessage.getSupportedVersions()
+
 export class PeerInfo {
     static newTracker(peerId: string, peerName?: string | null | undefined, location?: Location | null | undefined): PeerInfo {
-        return new PeerInfo(peerId, PeerType.Tracker, peerName, location)
+        return new PeerInfo(peerId, PeerType.Tracker, defaultControlLayerVersions, defaultMessageLayerVersions, peerName, location)
```
harbu: For completeness, these static helpers could take in `controlLayerVersions` and `messageLayerVersions`  parameters.

src/connection/WsEndpoint.ts
harbu: You can avoid having to use `// eslint-disable-next-line @typescript-eslint/no-unused-vars` by prefixing the unused parameter with underscore, i.e. `_peerId: string` in this case.

src/connection/WsEndpoint.ts
```diff
@@ -480,8 +495,18 @@ export class WsEndpoint extends EventEmitter implements IWsEndpoint {
         return this.peerBook.getAddress(peerId)
     }
 
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    getNegotiatedMessageLayerProtocolVersionOnNode(peerId: string): number | null {
+        return Math.max(...this.peerInfo.messageLayerVersions)
```
harbu: Wouldn't we want to take the maximum of the intersection of the two sets here?
```ts
const myVersions = this.peerInfo.messageLayerVersions
const theirVersions = this.peerBook.getByPeerId(peerId).messageLayerVersions
const sharedVersion = // intersection(myVersions, theirVersions)
return max(sharedVersions)
```

src/connection/WsEndpoint.ts
harbu: Same question as above

src/connection/WsEndpoint.ts
harbu: Hmm.. would it be easiest to just use `NegotiatedProtocolVersions` in this class (similar to `WebRtcEndpoint`)?

src/protocol/NodeToNode.ts
```diff
@@ -127,4 +128,12 @@ export class NodeToNode extends EventEmitter {
     getRtts(): Readonly<Rtts> {
         return this.endpoint.getRtts()
     }
+
+    getNegotiatedProtocolVersionsOnNode(peerId: string): number[] {
```
harbu: Since this method always returns a tuple, you can make the type signature tighter with: `[number, number]`.

src/protocol/NodeToNode.ts
```diff
@@ -127,4 +128,12 @@ export class NodeToNode extends EventEmitter {
...
+        const messageLayerVersion = this.endpoint.getNegotiatedMessageLayerProtocolVersionOnNode(peerId)
+            || Math.max(0, ...this.endpoint.getPeerInfo().messageLayerVersions)
```
harbu: Instead of `|| Math.max(0, ...this.endpoint.getPeerInfo().messageLayerVersions)` what about having a method on `NegotiatedProtocolVersions` for fetching the default version. That way you can also make sure it gets calculated at maximum once.

test/integration/node-to-node-protocol-version-negotiation.test.ts
```diff
@@ -0,0 +1,134 @@
...
+            host: '127.0.0.1',
+            port: 28680,
+            id: 'tracker'
+        })
+
+        const peerInfo1 = new PeerInfo('node-endpoint1', PeerType.Node, [1, 2, 3], [29, 30, 31])
+        const peerInfo2 = new PeerInfo('node-endpoint2', PeerType.Node, [1, 2], [31, 32, 33])
+        const peerInfo3 = new PeerInfo('node-endpoint3', PeerType.Node, [1, 2], [32])
+
+        // Need to set up TrackerNodes and WsEndpoint(s) to exchange RelayMessage(s) via tracker
```
harbu, line 37: :+1:

src/protocol/NodeToNode.ts
harbu: Hmm.. that got me thinking, should the entire logic of
```ts
return max(intersection(myVersions, theirVersions)) || max(myVersions)
```
be encapsulated in class `NegotiatedProtocolVersions`. This would require `WsEndpoint` to use the class however. But all the version calculation would be done there then?

test/integration/node-to-node-protocol-version-negotiation.test.ts
```diff
@@ -0,0 +1,134 @@
...
+            messageId: new MessageID('stream-1', 0, i, 0, 'node-endpoint1', 'msgChainId'),
+            prevMsgRef: null,
+            content: {
+                messageNo: i
+            },
+        })
+        nodeToNode1.sendData('node-endpoint2', msg1)
+    })
+
+    it('negotiated protocol is removed once node is disconnected', async () => {
```
harbu: negotiated <del>protocol</del> version

harbu (changes requested):
Looks good to me. A few thoughts that I think we should address.

### Version handling in `WsEndpoint`
Strictly speaking, we don't need version handling in `WsEndpoint` since this ticket was concentrated on versions between two nodes, not node and tracker. That being said, it is easy to see that in the future we may want to also negotiate versions between node and tracker. Furthermore, if we want to keep around the possibility of using `WsEndpoint` between nodes, having version negotiation on `WsEndpoint` is a must.

Could we use the `NegotiatedProtocolVersion` class in `WsEndpoint` similarly to how is done in `WebRtcEndpoint`? Then we could delegate all calculations to that class and have it be the single source of truth.


### Default version in the presence of no version information
To my understanding, the code works so that if no protocol  version information is supplied we assume the latest versions of protocols are used. Wondering if it should just be an error not to supply version information? Argument being that you catch the erroneous situation earlier vs. receiving latest version messages.

### Unit test for NegotiatedProtocolVersions
Although it gets tested indirectly by `node-to-node-protocol-version-negotiation.test.ts`, the class could use a unit test of its own covering edge cases etc. Generally just makes it quicker to spot any errors when making e.g. changes to the class.

src/connection/WsEndpoint.ts
juslesan: I only added these functions to the WS endpoint to ensure that the function exists in both endpoint types. The WS negotiated protocol is not actually used anywhere. We do still use WS for node-to-node connections in one test: `integration/delivery-of-messages-protocol-layer.test.ts` which is why I had to include the function. Maybe the WS endpoints should also fully negotiate the used protocol to ensure that WS node-to-node connections will work if we end up bringing them back.

src/connection/WsEndpoint.ts
juslesan: > I only added these functions to the WS endpoint to ensure that the function exists in both endpoint types. The WS negotiated protocol is not actually used anywhere. We do still use WS for node-to-node connections in one test: integration/delivery-of-messages-protocol-layer.test.ts which is why I had to include the function. Maybe the WS endpoints should also fully negotiate the used protocol to ensure that WS node-to-node connections will work if we end up bringing them back.

src/connection/WebRtcEndpoint.ts
juslesan: Currently there should only ever be a single connection per peer

juslesan:
> Strictly speaking, we don't need version handling in `WsEndpoint` since this ticket was concentrated on versions between two nodes, not node and tracker.

This is true. I included exchanging information of protocol versions on WSEndpoint as it will turn out to be useful in the future. For example, the tracker algorithm could take in to account that some nodes cannot connect to eaach other if they do not share protocol versions.  Alternatively, the tracker could reject WS connections from nodes that are too far behind in the versioning. That being said, I'm fine with the logic being removed for now if necessary.

src/protocol/NodeToNode.ts
timoxley: Was thinking the same thing: https://github.com/streamr-dev/network/pull/556#discussion_r628480431

harbu (approved):
LGTM

PR #558, opened 2021-05-10
NET-261: Improve logging

teogeb:
Improvements for `Logger.ts`:

- get name from NodeJS `module`
- prefix with `STREAMR_APPLICATION_ID` environment variable
- remove colorizing (as Broker will colorize using the application id)
- simplify output

New helper class `NameDirectory` to get a human-readable name of Broker/Tracker.  Also `TrackerServer.send` prints human-readable message type name.

Changed log level for some classes from `debug` to `trace`.


src/helpers/Logger.ts
```diff
@@ -1,39 +1,38 @@
...
+    constructor(module: NodeJS.Module, context?: string) {
         this.logger = pino({
-            name: formName(path, peerInfo),
+            name: Logger.createName(module, context),
             enabled: !process.env.NOLOG,
             level: process.env.LOG_LEVEL || 'info',
             prettyPrint: process.env.NODE_ENV === 'production' ? false : {
-                colorize: true,
-                translateTime: true
+                colorize: false,
```
harbu: I think it would be useful to enable colorized logs for network too. It helps out when running the test suite and trying to visually identify who logged what.

src/helpers/Logger.ts
```diff
@@ -1,39 +1,38 @@
...
         })
     }
 
-    createChildLogger(subPath: string[]): Logger {
-        return new Logger([...this.path, ...subPath], this.peerInfo)
+    private static createName(module: NodeJS.Module, context?: string) {
+        const parsedPath = path.parse(module.filename)
+        let fileId = parsedPath.name
+        if (fileId === 'index') {
+            // file with name "foobar/index.ts" -> "foobar"
```
harbu, line 42: Good clarifying comment 👍

src/helpers/Logger.ts
```diff
@@ -1,39 +1,38 @@
...
+            const parts = parsedPath.dir.split(path.sep)
+            fileId = parts[parts.length - 1]
+        }
+        const appId = process.env.STREAMR_APPLICATION_ID
+        const longName = _.without([appId, fileId, context], undefined).join(':')
```
harbu, line 47: Unclear to me what `._without` does here. Based on documentation https://www.geeksforgeeks.org/underscore-js-_-without-with-examples/ the 2nd argument takes a list of values to remove from the first argument array, but here the it is undefined?

src/helpers/Logger.ts
```diff
@@ -1,39 +1,38 @@
...
+                translateTime: 'yyyy-mm-dd"T"HH:MM:ss.l',
+                ignore: 'pid,hostname',
+                levelFirst: true,
             }
         })
     }
 
-    createChildLogger(subPath: string[]): Logger {
-        return new Logger([...this.path, ...subPath], this.peerInfo)
+    private static createName(module: NodeJS.Module, context?: string) {
```
harbu, line 38: Considering writing a unit test for this method.

src/helpers/Logger.ts
```diff
@@ -59,4 +58,11 @@ export class Logger {
     trace(msg: string, ...args: any[]): void {
         this.logger.trace(msg, ...args)
     }
+    
+    getFinalLogger() {
```
harbu: return type

src/helpers/messageType.ts
```diff
@@ -0,0 +1,41 @@
+import ControlMessage from 'streamr-client-protocol/dist/src/protocol/control_layer/ControlMessage'
+import TrackerMessage from 'streamr-client-protocol/dist/src/protocol/tracker_layer/TrackerMessage'
+
+// TODO move this to protocol-js?
```
harbu: Would be better in protocol-js, and if possible, even have the protocol messages themselves read their versions from these mappings. Think there is a high chance to forget to update the mapping here.

src/protocol/TrackerNode.ts
```diff
@@ -139,6 +140,7 @@ export class TrackerNode extends EventEmitter {
     }
 
     onPeerConnected(peerInfo: PeerInfo): void {
+        this.logger.debug(`Peer connected: ${NameDirectory.getName(peerInfo.peerId)}`)
```
harbu, line 143: Subjective, personally I prefer printf-style logging. (I.e. "Peer connected %s", ...)

harbu (commented):
Overall LGTM; few small things. Maybe the biggest question regarding message type listing and whether it should be in protocol?

src/helpers/Logger.ts
teogeb: We could add an environment variable for this (Broker's `run-all.js` would disable colorizing by setting that variable)

teogeb: Added environment variable

src/helpers/Logger.ts
teogeb: It is a variadic function: the first argument is the source array and all other parameters are items which it removes from the source array. In this case we remove all undefined values from the list (in practice `appId` and/or `context`).
```
const items = ['foo', 'bar', undefined, 'ipsum']
console.log(_.without(items, 'bar', 'foo')) // [ undefined, 'ipsum' ]
console.log(_.without(items, undefined)) // [ 'foo', 'bar', 'ipsum' ]
```

src/helpers/Logger.ts
teogeb: Added

src/helpers/Logger.ts
teogeb: Fixed

src/helpers/messageType.ts
teogeb: Removed the utility, we can read the values directly from enums. Needs this PR: https://github.com/streamr-dev/streamr-client-protocol-js/pull/64

src/helpers/Logger.ts
harbu: Aaaah I get it now.

harbu (approved):
LGTM!

PR #562, opened 2021-05-20
Rework reconnects, fix flaky tests

juslesan:
Mostly fixes flaky tests in CI.

Fast webRTC reconnect test is still sometimes fails in CI, but the occurances are rare. It appears that in rare cases a connection is closed just before the RemoteDescription is being set, this causes the connection forming process to stall. This could be fixed with a queuing mechanism, but expecting all such cases where the connection is closed and offer / answer signalling messages are received immediately after to wait for reconnects might lead to weird behaviour.

- Remove force connect logic
- Non-offerer always uses rtcConnect
- If connections have set RemoteDescription any new rtcConnect and rtcOffer messages are interpreted as reconnect attempts
- RemoteCandidate can no longer be called before RemoteDescription has been set
- Rework 'Messages are delivered on temporary loss of connection' test case


src/connection/WebRtcEndpoint.ts
```diff
@@ -75,8 +75,14 @@ export class WebRtcEndpoint extends EventEmitter implements IWebRtcEndpoint {
             })
             const connection = this.connections[peerId]
             if (connection) {
+                if (connection.isRemoteDescriptionSet()) {
+                    this.close(peerId, 'rtcOffer message received for a new connection')
+                    this.connect(peerId, routerId)
```
harbu: `catch` (or await?) for `this.connect`

test/unit/WebRtcEndpoint.test.ts
```diff
@@ -129,6 +136,15 @@ describe('WebRtcEndpoint', () => {
         await waitForCondition(() => ep2NumOfReceivedMessages === 10)
     })
 
+    it('connection is formed when only non-offerer invokes connect()', async () => {
```
harbu, line 139: :+1:

harbu (approved):
LGTM! (I added a small comment to code to explain the offer after remoteDescription thing)

